Issue #1, opened 2017-08-22
The SELinux role does not enable SELinux if disabled.


pcahyna:
`/etc/selinux/config` contains:
```
SELINUX=disabled
SELINUXTYPE=targeted 
```
SELinux is disabled.
When I run `ansible-playbook -c local -i localhost, selinux-playbook.yml`, where `selinux-playbook.yml` has
```yaml
---
- hosts: all
  become: yes
  become_method: sudo
  become_user: root
  vars:
    SELinux_type: targeted
    SELinux_mode: enforcing
    SELinux_change_running: 1

    SELinux_booleans:
      - { name: 'samba_enable_home_dirs', state: 'on' }
      - { name: 'ssh_sysadm_login', state: 'on', persistent: 'yes' }

  roles:
    - rhel-system-roles.selinux
```
I get the following error:
```
PLAY [all] *********************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [rhel-system-roles.selinux : Install SELinux tools] ***********************
ok: [localhost]

TASK [rhel-system-roles.selinux : Install SELinux tool semanage on Fedora] *****
skipping: [localhost]

TASK [rhel-system-roles.selinux : Drop all local modifications first] **********
[...]

PR #2, opened 2017-08-24
Allow setting SELinux mode when SELinux is disabled

bachradsusi:
When SELinux is disabled some operations like policy reload can't be
done therefore we need to:

- Change the permanent SELinux mode first

- Use '-N' option in 'semanage' import commands in order to avoid
automatic policy reloading

- Reload policy only if SELinux is not disabled

The new order of steps used in the role is:

1. Install SELinux tools
2. Set permanent SELinux mode via /etc/selinux/config
3. Set running SELinux mode - works only when SELinux is not disabled
4. Drop all local modifications - this is the place where '-N' option is added so that it works with SELinux disabled
5. Reload SELinux policy - only if SELinux is not disabled
6. Set SELinux file contexts, booleans and so on - all of them don't work with SELinux disabled due to limitations in Ansible modules

Fixes: https://github.com/linux-system-roles/selinux/issues/1

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,9 +14,6 @@
     state: present
   when: ansible_distribution == "Fedora"
 
-- name: Drop all local modifications first
-  shell: echo "{{drop_local_modifications}}" | /usr/sbin/semanage -i -
-
 - name: Set permanent SELinux mode
   selinux: policy={{ SELinux_type }} state={{ SELinux_mode }}
   when: SELinux_mode is defined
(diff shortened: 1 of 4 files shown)


larskarlitski:
Thanks! Do you mind adding a test for the new behavior? Seems like that will be easy to break again in the future.

bachradsusi:
Sure. I'll do a test and update this PR.

bachradsusi:
Commits related to new test were merged into one.

The test tries to simulate SELinux disabled system using `umount <selinuxfs>` where `selinuxfs` is detected using `findmnt -n -t selinuxfs --output=target` which should work on supported systems - Red Hat Enterprise Linux 6 and 7 and recent Fedora's

larskarlitski:
Thanks!

Issue #3, opened 2017-08-29
`test_selinux_disabled` fails on centos-6


larskarlitski:
```
TASK [selinux : Drop all local modifications] **********************************
task path: /role/tasks/main.yml:25
Using module file /usr/lib/python2.7/site-packages/ansible/modules/core/commands/command.py
<localhost> ESTABLISH SSH CONNECTION FOR USER: admin
<localhost> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o Port=2222 -o 'IdentityFile="/var/tmp/avocado_zrDGNN/3-test.py_Test.test_3/tmpEDB8PZ/id_rsa"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=admin -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ControlPath=/home/tester/.ansible/cp/ansible-ssh-%h-%p-%r localhost '/bin/sh -c '"'"'( umask 77 && mkdir -p "` echo ~/.ansible/tmp/ansible-tmp-1504016091.84-192915026227876 `" && echo ansible-tmp-1504016091.84-192915026227876="` echo ~/.ansible/tmp/ansible-tmp-1504016091.84-192915026227876 `" ) && sleep 0'"'"''
<localhost> PUT /var/tmp/tmp5g6jvq TO /home/admin/.ansible/tmp/ansible-tmp-1504016091.84-192915026227876/command.py
<localhost> SSH: EXEC sftp -b - -C -o ControlMaster=auto -o ControlPersist=60s -o Port=2222 -o 'IdentityFile="/var/tmp/avocado_zrDGNN/3-test.py_Test.test_3/tmpEDB8PZ/id_rsa"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=admin -o ConnectTimeout=10 -o ControlPath=/home/tester/.ansible/cp/ansible-ssh-%h-%p-%r '[localhost]'
<localhost> ESTABLISH SSH CONNECTION FOR USER: admin
[...]

PR #4, opened 2017-09-05
Fix policy reload problem on CentOS 6 and update tests

bachradsusi:
'semanage -i -' command CentOS 6 and Red Hat Enterprise Linux 6 has a
bug which resets 'noreload' global variable every time a new line is
read even when the line is empty. E.g. for

echo "fcontext -D -N\n" | semanage -i -

Since echo adds another '\n' at the end of the string it means that the
empty line at the end of the input string changes the value of
'noreload' to True and semanage tries to reload policy and it fail as
SELinux is disabled. Note: while the variable is called 'noreload' it
has the opposite meaning.

Fixes: #3

At the same time, tests are being updated to cleanup when pass and  `test_selinux_disabled` has been enhanced to cover issue #3

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -23,7 +23,7 @@
   when: SELinux_mode is defined and SELinux_change_running is defined
 
 - name: Drop all local modifications
-  shell: echo "{{drop_local_modifications}}" | /usr/sbin/semanage -i -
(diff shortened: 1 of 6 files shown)


bachradsusi:
I ran test locally several times without any problem. I'll try to find out why it failed if I did something wrong.

bachradsusi:
I've updated the tests and everything passes now.

PR #5, opened 2017-09-20
tests to test semaphore tests on CentOS 6  - do not merge :)

bachradsusi:
This is a testing pull request which should fail on CentOS 6. It's based on https://github.com/linux-system-roles/selinux/pull/4 but without the real fix. There are only changes in tests.

Do not merge

### DIFF
--- a/test/test_boolean.yml
+++ b/test/test_boolean.yml
@@ -5,8 +5,7 @@
 
   vars:
     SELinux_booleans:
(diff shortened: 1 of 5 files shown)


bachradsusi:
This was requested by @pcahyna to confirm the fix from the other pull request

larskarlitski:
Can we close this, then?

pcahyna:
Yes please.

PR #11, opened 2018-07-02
Remove notes about bug in fcontext ansible module.

wrabcak:
https://bugzilla.redhat.com/show_bug.cgi?id=1405110 is fixed in current
release.

Tested with:
ansible-2.5.5-1.el7.noarch


pcahyna:
good catch, thanks

PR #12, opened 2018-07-18
Validate that selinux_policy is present

pcahyna:
and fail with a better error message if not, but do not require it when state is disabled.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,8 +14,13 @@
     state: present
   when: ansible_distribution == "Fedora"
 
+- name: validate selinux_policy
+  fail:
+    msg: selinux_policy must be provided if selinux_state is not "disabled"
+  when: selinux_state != "disabled" and selinux_policy is not defined
+
 - name: Set permanent SELinux state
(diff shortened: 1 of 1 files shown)


bachradsusi:
Would you please explain what kind of problem this PR tries to fix?

pcahyna:

pcahyna:
> What would happen if both selinux_state and selinux_policy were not defined?

Bad things would happen. See #13 to fix them.

#15 includes both changes.

> Could selinux_policy be defined in defaults/main.yml as targeted which is expected to be installed and used in most cases?

This is a valid option as well. My only concern is that the options would be inconsistent (`selinux_state` would be required but `selinux_policy` would have a default).

bachradsusi:
I don't think that `selinux_state` should be always required. You don't have to change the state every time you want to add/change/remove some configuration.

wrabcak:
Agree with @bachradsusi . 
Closing this PR. We'll create new PR where selinux_policy will be in defaults with value targeted.

PR #13, opened 2018-07-19
Validate that selinux_state is present

pcahyna:
Fail with an error message if not.

PR #15, opened 2018-07-19
Validate that selinux_policy is present

pcahyna:
Includes commits from both #12 and #13


wrabcak:
Same situation like #12 . Closing.

PR #16, opened 2018-07-19
Do not override local modifications by default

wrabcak:
Following PR resolves: 
https://bugzilla.redhat.com/show_bug.cgi?id=1597545

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,20 @@ roles:
     become: true
 ```
 
+#### purge local modifications using appropriate variable
+
+```yaml
+selinux_booleans_purge: true
+selinux_fcontexts_purge: true
+selinux_ports_purge: true
+selinux_logins_purge: true
+```
+
+#### purge all local modifications using variable
+
+```yaml
+selinux_all_purge: true
+```
 
 #### set SELinux policy type and mode
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -47,7 +47,7 @@
     target: "{{ item.target }}"
     setype: "{{ item.setype }}"
     ftype:  "{{ item.ftype | default('a') }}"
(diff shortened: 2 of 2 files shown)


bachradsusi:
I have only cosmetic issues with PR:

I would squash d407864 and d9921ca together.

`selinux_all_purge` is not documented in commit message and none of new variables is documented in README.

It could be useful to have also tests to cover new variables and new functionality and to be sure that we don't override system modifications.

Otherwise it looks good.

wrabcak:
> I would squash d407864 and d9921ca together.
Fixed.

> selinux_all_purge is not documented in commit message and none of new variables is documented in README.
Fixed.

> It could be useful to have also tests to cover new variables and new functionality and to be sure that we don't override system modifications.

Let's create issue for missing tests.

README.md
```diff
@@ -45,6 +45,15 @@ roles:
...
 
+#### purge local modifications using appropriate variable
+
+```yaml
+selinux_all_purge: 1
+selinux_booleans_purge: 1
+selinux_fcontexts_purge: 1
+selinux_ports_purge: 1
+selinux_logins_purge: 1
+```
```
bachradsusi: There were some comments about using boolean True/False instead of numbers. Should this be the case?

wrabcak: Fixed.

pcahyna:
[citest bad]

README.md
pcahyna: having `selinux_all_purge: true` with the other `*purge` variables in the same yaml example is IMO confusing. Can you split it into two example files, one with `selinux_all_purge`, the other with the others?

wrabcak:
Looks like this PR is ready. Will rebase and merge.

PR #17, opened 2018-07-19
Add default value "targeted" for selinux_policy variable

wrabcak:
Resolves:

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,0 +1,1 @@
+selinux_policy: targeted


bachradsusi:
Could you please rebase your branch to the latest master and repush?

selinux-playbook.yml
```diff
@@ -4,7 +4,6 @@
   become_method: sudo
   become_user: root
   vars:
-    selinux_policy: targeted
```
bachradsusi: I'd leave in the example playbook as it is so it's clean that you can override it.

wrabcak: Make sense, will change it.

wrabcak:
We find better solution like this.  We'll take the current value from ansible facts if the variable is not defined.

PR #19, opened 2018-07-19
Handle selinux_state and selinux_policy variables if there are not defined.

wrabcak:
If these variables are not defined, will be used current value from
Ansible facts from ansible node.

- selinux_state using ansible_selinux.config_mode
- selinux_policy using ansible_selinux.type

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1602901

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -24,9 +24,21 @@
     ( ansible_distribution_major_version > "7" and
       ( ansible_distribution == "CentOS" or ansible_distribution == "RedHat" ))
 
-- name: Set permanent SELinux state
-  selinux: policy={{ selinux_policy }} state={{ selinux_state }}
-  when: selinux_state is defined
+- name: Set permanent SELinux state if enabled
+  selinux:
+    state: "{{ selinux_state | default(ansible_selinux.config_mode) }}"
+    policy: "{{ selinux_policy | default(ansible_selinux.type) }}"
+  when: ansible_selinux.status == "enabled"
+
+- name: Set permanent SELinux state if disabled
+  selinux:
+    state: "{{ selinux_state }}"
+    policy: "{{ selinux_policy | default('targeted') }}"
+  when: ansible_selinux.status == "disabled" and selinux_state is defined
+
+- debug:
+    msg: "SELinux is disabled on system - some SELinux modules can crash"
+  when: ansible_selinux.status == "disabled"
 
 - name: Drop all local modifications
   shell: echo -e -n "{{drop_local_modifications}}" | /usr/sbin/semanage -i -


tasks/main.yml
```diff
@@ -24,9 +24,21 @@
...
+
+- name: Set permanent SELinux state is disabled
+  selinux:
+    state: "{{ selinux_state }}"
+    policy: "{{ selinux_policy | default('targeted') }}"
+  when: ansible_selinux.status == "disabled" and selinux_state is defined
+
+- debug:
+    msg: "SELinux is disabled on system and current playbook doesn't enabling it, some SELinux modules can crash"
+  when: ansible_selinux.status == "disabled" and selinux_state is not defined
```
pcahyna: "does not enable it"

tasks/main.yml
```diff
@@ -24,9 +24,21 @@
...
       ( ansible_distribution == "CentOS" or ansible_distribution == "RedHat" ))
 
-- name: Set permanent SELinux state
-  selinux: policy={{ selinux_policy }} state={{ selinux_state }}
-  when: selinux_state is defined
+- name: Set permanent SELinux state if enabled
+  selinux:
+    state: "{{ selinux_state | default(ansible_selinux.config_mode) }}"
+    policy: "{{ selinux_policy | default(ansible_selinux.type) }}"
+  when: ansible_selinux.status == "enabled"
```
bachradsusi: If SELinux is enabled, `selinux` will be always run even when it does not change anything because it uses current values.

What about:

`when: ansible_selinux.status == "enabled" and ( selinux_state is defined or selinux_state is defined )`

tasks/main.yml
```diff
@@ -24,9 +24,21 @@
...
+
+- name: Set permanent SELinux state is disabled
```
bachradsusi: `s/is/if/`

tasks/main.yml
```diff
@@ -24,9 +24,21 @@
...
+  selinux:
+    state: "{{ selinux_state }}"
+    policy: "{{ selinux_policy | default('targeted') }}"
+  when: ansible_selinux.status == "disabled" and selinux_state is defined
+
+- debug:
+    msg: "SELinux is disabled on system and current playbook does not enable it, some SELinux modules can crash"
```
bachradsusi: I thing this message needs to be reformulated. The playbook could enable it using `selinux_state` in task before but it would take effect after reboot.

wrabcak: Sorry, Don't understand what you mean. Could you add example how it should be reformulated?

bachradsusi: "SELinux is disabled on system - some SELinux modules can crash"  would be enough. The part "current playbook does not enable it" could be confusing when playbook sets `selinux_state` to `enforcing`

wrabcak: Fixed.

tasks/main.yml
wrabcak: Fixed

tasks/main.yml
wrabcak: Fixed

pcahyna: > If SELinux is enabled, selinux will be always run even when it does not change anything because it uses current values.

Not entirely true. It uses current _configured_ values (`config_mode` which can be different from the running configuration (`mode`).

tasks/main.yml
pcahyna: >  The playbook could enable it using selinux_state in task before but it would take effect after reboot.

In that case we will fail earlier and don't reach this message, do we?

tasks/main.yml
pcahyna: @bachradsusi I actually think you were right and one should not call the module when not `( selinux_state is defined or selinux_policy is defined )`. Not because it would not change anything (that would be an unimportant optimization), but because it could change something (reset the running state to the saved state if they differ) and such change would be unexpected if one has not defined `selinux_state` or `selinux_policy`.

I filed #24 to return to your proposal (with @wrabcak 's correction).

PR #20, opened 2018-07-19
define the behavior of the role when SELinux is disabled

wrabcak:
Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1597576


bachradsusi:
It could be useful to have `selinux_reboot_required` fact documented in README

wrabcak:
Documentation of `selinux_reboot_required` fact added.

tasks/main.yml
```diff
@@ -24,9 +24,32 @@
...
+    state: "{{ selinux_state | default(ansible_selinux.config_mode) }}"
+    policy: "{{ selinux_policy | default(ansible_selinux.type) }}"
+  register: selinux_mod_output
+  when: ansible_selinux.status == "enabled"
+
+- name: Set permanent SELinux state is disabled
+  selinux:
+    state: "{{ selinux_state }}"
+    policy: "{{ selinux_policy | default('targeted') }}"
+  register: selinux_mod_output
```
pcahyna: This will not work as the second selinux task will clobber the `selinux_mod_output` result from the first one.

README.md
```diff
@@ -82,3 +82,9 @@ selinux_restore_dirs:
       - { login: 'plautrba', seuser: 'staff_u', state: 'absent' }
       - { login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023', state: 'present' }
 ```
+
+## Ansible Facts
+
+### selinux\_reboot\_required
+
+This custom fact is set to `true` if system reboot is necessary when SELinux is set from `disabled` to `enabled` or vice versa.  Otherwise `false` is set to the fact.
```
pcahyna: ... the fact is set to `false`.

wrabcak: Fixed.

PR #22, opened 2018-07-23
Update tests for SELinux role.

wrabcak:
selinux_*_purge variables was added, which means you need to specify
right variable to drop local modifications. This patch fixing our tests.


pcahyna:
LGTM, please merge!

pcahyna:
The tests seem insufficient. They don't really check that previous local modifications were purged, or do they? Also, it is be needed to set *_purge to no and test that local modifications are **not** dropped.

PR #25, opened 2018-07-24
Fix tests_selinux_disabled.yml

pcahyna:
The role nowadays fails if it finds out that the machine needs to be rebooted.
Adapt the test to this and check that the role has indeed failed and has set the correct variable.

Other minor corrections.


wrabcak:
LGTM.

PR #26, opened 2018-07-24
Mention that the role fails when reboot is needed

pcahyna:
and document how to handle it. Other minor fixes to the example playbook.


wrabcak:
LGTM.

PR #29, opened 2018-07-31
Update tests after the recent change in semantics

pcahyna:
test all state transitions, test that the role does not change state by default and other test updates.


pcahyna:
In particular this tests #19 and #20

pcahyna:
[citest bad]

pcahyna:
[citest bad]

pcahyna:
[citest bad]

pcahyna:
[citest bad]

wrabcak:
@pcahyna, 

CI still failing, could you check it please? 

Thanks,
Lukas.

pcahyna:
They are not failing, they encounter an error during test VM setup due to some problem in the test harness:
ERROR: Inventory not properly set up
Running them repeatedly usually makes them succeed (probably because the result depends on what machine in the test cluster executes the test).

tyll:
The status contains an identifier such as `test-harness-f28-38-9t8bh`. The last part is the pod's ID. If they fail because of /dev/kvm permissions it helps to delete the pod and hope that the new pod will be created on a different node. I tried this now.

pcahyna:
[citest bad]

pcahyna:
@wrabcak after several retries the tests pass without code changes.

PR #30, opened 2018-08-15
Update documentation

yontalcar:


README.md
```diff
@@ -67,6 +69,8 @@ selinux_policy: targeted
 selinux_state: enforcing
 ```
```
pcahyna: It should be said that if `selinux_state` is not given, it is not changed (the previous value is preserved).

yontalcar: amended

PR #33, opened 2018-10-01
Move defaults to defaults/main.yml

major:
Signed-off-by: Major Hayden <major@redhat.com>

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,0 +1,15 @@
+---
+
+# Set up empty lists for SELinux changes.
+selinux_booleans: []
+selinux_fcontexts: []
+selinux_logins: []
+selinux_ports: []
+selinux_restore_dirs: []
+
+# Purging local modifications is disabled by default.
+selinux_all_purge: no
+selinux_booleans_purge: no
+selinux_fcontexts_purge: no
+selinux_ports_purge: no
+selinux_logins_purge: no
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -59,23 +59,23 @@
 
 - name: Drop all local modifications
   shell: echo -e -n "{{drop_local_modifications}}" | /usr/sbin/semanage -i -
-  when: selinux_all_purge | default(False)
+  when: selinux_all_purge | bool
 
 - name: Purge all SELinux boolean local modifications
   shell: /usr/sbin/semanage boolean -D
-  when: selinux_booleans_purge | default(False)
+  when: selinux_booleans_purge | bool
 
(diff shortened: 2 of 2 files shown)


pcahyna:
[citest commit:41a866de8bbd5be54575de5a2f30e013333003ac]

wrabcak:
@pcahyna 
Where is CI? PR looks good but I would like to see tests passing. :) 

Thanks,
Lukas.

pcahyna:
I don't know, there seems to be some problem with the CI. @tyll , do you know anything?

tyll:
@pcahyna according to the GitHub API you are only considered to be a CONTRIBUTOR and not a MEMBER, therefore the CI ignores your comment. @wrabcak is considered to be a MEMBER.

tyll:
[citest commit:41a866de8bbd5be54575de5a2f30e013333003ac]

It seems that I am also a MEMBER, let's check...

pcahyna:
By the way, our CI did not run for https://github.com/linux-system-roles/network/pull/86 either.

pcahyna:
@wrabcak here are your Ci results, can you merge the PR?

wrabcak:
LGTM. 

Thanks

pcahyna:
@yontalcar maybe use this API https://developer.github.com/v3/repos/collaborators/#review-a-users-permission-level to find out whether the user has push permissions? (Instead of checking `comment["author_association"] == "MEMBER"` https://github.com/linux-system-roles/test-harness/blob/5801b7fae06cb96818fc9308f340aa01d93b4be4/test/run-tests#L325 )

tyll:
@pcahyna I took another look. It seems that your membership in the system roles organization is not public:
https://github.com/orgs/linux-system-roles/people

Therefore you appear as a contributor instead of a member in some cases. For example when looking at the comments in this PR with an unauthenticated browser (for example using the private mode), you are mentioned to be a contributor. But when I look after authentication, it shows you as a member. I recall that I read about the public membership to be required for external services as well.

yontalcar:
@tyll: Looks like adding `read:org` scope to token used for `test-harness` should be sufficient to fix this.

pcahyna:
@yontalcar great! Can you please create another (fake) PR which illustrates the problem and then we can try changing the token?
(Actually the PR should be created by somebody who is not the member of the organization probably to reproduce what happened here)

pcahyna:
@tyll 
> I took another look. It seems that your membership in the system roles organization is not public

It's not just me. It seems that almost all members' membership is private. Anyway, thanks for discovering this.

PR #35, opened 2018-10-18
Add support for selinux_permissive module

shamsalmon:
Added support for turning off/on permissive mode for specific domains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,19 @@ selinux_booleans:
   - { name: 'ssh_sysadm_login', state: 'on', persistent: 'yes' }
 ```
 
+#### set SELinux domains
+
+```yaml
+selinux_domains:
+  - { name: 'httpd_t', state: 'permissive' }
+```
+
+```yaml
+selinux_booleans:
+  - { name: 'http', state: 'on' }
+  - { name: 'ssh_sysadm_login', state: 'on', persistent: 'yes' }
+```
+
 #### Set SELinux file contexts
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -86,13 +86,7 @@ selinux_booleans:
 
 ```yaml
 selinux_domains:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -82,6 +82,19 @@ selinux_booleans:
...
+selinux_domains:
+  - { name: 'httpd_t', state: 'permissive' }
+```
+
+```yaml
+selinux_booleans:
+  - { name: 'http', state: 'on' }
+  - { name: 'ssh_sysadm_login', state: 'on', persistent: 'yes' }
+```
+
```
bachradsusi: It looks like copy&paste leftover. selinux_booleans are shown above this chapter.

bachradsusi:
I'd rather use something like `selinux_permissive_domains` so it's clear that it handles just permissive domains.

pcahyna:
> I'd rather use something like `selinux_permissive_domains` so it's clear that it handles just permissive domains.

An alternative would be to keep `selinux_domains` and use it for any other domain attributes that may arise in the future, if such thing is likely to arise.

shamsalmon:
Ill make those changes guys. Thanks!

shamsalmon:
@bachradsusi made some suggested changes. Let me know if you guys would like anything else!

bachradsusi:
I'm fine with this change generally. Could you please provide also a test for the `selinux_domains` ?

pcahyna:
[citest commit:af5a0a64230c78bd9154111a4287064f1f96da59]

pcahyna:
@shamsalmon @bachradsusi what do you think about the proposed `selinux_domains_purge` option? I would like to add it for symmetry with others, unless it makes no sense.

bachradsusi:
>  what do you think about the proposed selinux_domains_purge option?

Given that there's no `semanage permissive -D` yet I'm not sure it's worth to implement it until it's supported directly in `selinux_permissive` module

PR #43, opened 2019-05-14
List all variables in defaults.

pcahyna:
Those whose absence had a special meaning have a default value of null.

The second argument to the default() Jinjs2 filter was used to compensate for
the change: it causes the variable to be substituted if evaluates to false
(instead of just if undefined).

This helps using the role in Foreman/Satellite, because those introspect the
role's defaults file to know what parameters the role takes and present it to
the user. (So far, there has been no other way to declare role's input
parameters.)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,4 +1,6 @@
 ---
+selinux_state: null
+selinux_policy: null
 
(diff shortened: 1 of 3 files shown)


pcahyna:
[citest bad]

PR #44, opened 2019-06-12
Add cleanup to tests_selinux_disabled.yml test. Tag tests as Tier1.

vcrhonek:


tests/tests_selinux_disabled.yml
```diff
@@ -81,3 +82,10 @@
         name: sar-user
         remove: yes
         state: absent
```
pcahyna: I think those three tasks above should be also tagged as `[ 'tests::cleanup', 'tests::tier1::cleanup' ]`. Maybe use a block?

pcahyna: Applies also to "Backup original /etc/selinux/config", one does not need to back it up, if one does not need to restore it.

vcrhonek: Fixed.

tyll:
[citest bad]

richm:
@vcrhonek @pcahyna do we still want this PR?  We are carrying a patch downstream that is very similar to this PR.  Why does this use `import_tasks` instead of `include_tasks`?

richm:
closing in favor of https://github.com/linux-system-roles/selinux/pull/77

PR #51, opened 2020-08-31
collections prep - use FQRN

richm:
Prepare for collections - use fully qualified role name


nhosoi:
` tests/roles/linux-system-roles.selinux --> ../../` ?

richm:
> ` tests/roles/linux-system-roles.selinux --> ../../` ?

fixed in https://github.com/linux-system-roles/selinux/pull/51/commits/892b44014fb3129e466065fc05b2d730b03640eb

PR #55, opened 2020-10-20
Drop selogin module

bachradsusi:
selogin module is available directly via Ansible since 2.8, later
migrated to community.general collection.

https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/modules/system/selogin.py
https://github.com/ansible-collections/community.general/blob/main/plugins/modules/selogin.py


richm:
please rebase

jharuda:
[citool]

richm:
@bachradsusi do you want to get this into 8.5?  Can you rebase?

bachradsusi:
Rebased. But I wouldn't mind if this is dropped. It depends how this and other Ansible related projects handle situation when an originally downstream only module was upstreamed and available from collections.

richm:
> Rebased. But I wouldn't mind if this is dropped. It depends how this and other Ansible related projects handle situation when an originally downstream only module was upstreamed and available from collections.

I think we (system roles team, selinux team, ansible team) are going to have to figure out a way to get the modules used by this role into a supported collection.  `selinux` and `seboolean` are in the supported posix collection https://docs.ansible.com/ansible/latest/collections/ansible/posix/index.html#plugins-in-ansible-posix - so one option would be to get `sefcontext` and `selogin` and `seport` and `selinux_permissive` into `ansible.posix` - it makes sense that all of the selinux related modules are in the same collection.

richm:
@bachradsusi can you also remove .sanity-ansible-ignore-2.10.txt  and .sanity-ansible-ignore-2.9.txt in this PR since they are no longer useful without selogin.py?

bachradsusi:
> @bachradsusi can you also remove .sanity-ansible-ignore-2.10.txt and .sanity-ansible-ignore-2.9.txt in this PR since they are no longer useful without selogin.py?

removed

PR #68, opened 2021-02-11
use tox-lsr 2.2.0

nhosoi:
Upgrade to tox-lsr 2.2.0 to pick up fix for improving collection
testing to get rid of black.


richm (approved):
lgtm

PR #69, opened 2021-02-23
use reboot module; ansible 2.8

richm:
Use the `reboot` module to restart the machine and wait for it to
come back, rather than `shell` and a separate wait task.
Change `min_ansible_version` to 2.8

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -10,7 +10,7 @@ galaxy_info:
     - centos
   company: Red Hat, Inc.
   license: GPL-3.0+
-  min_ansible_version: 2.5
+  min_ansible_version: 2.8
   platforms:
     - name: Fedora
       versions:
--- a/tests/selinux_apply_reboot.yml
+++ b/tests/selinux_apply_reboot.yml
@@ -10,12 +10,8 @@
       when: not selinux_reboot_required
 
     - name: restart managed host
-      shell: sleep 2 && shutdown -r now "Ansible updates triggered"
-      async: 1
-      poll: 0
-      ignore_errors: true
-
-    - name: wait for managed host to come back
-      wait_for_connection:
-        delay: 10
-        timeout: 300
+      reboot:
+        msg: Ansible updates triggered
+        pre_reboot_delay: 2
+        post_reboot_delay: 10
+        reboot_timeout: 300


tests/selinux_apply_reboot.yml
```diff
@@ -10,12 +10,7 @@
...
-      ignore_errors: true
-
-    - name: wait for managed host to come back
-      wait_for_connection:
-        delay: 10
-        timeout: 300
+      reboot:
+        msg: Ansible updates triggered
+        pre_reboot_delay: 2
+        reboot_timeout: 3000
```
spetrosi, line 16: 3000 seconds = 50 minutes, is this intentional?

richm: good catch - I thought delay was the number of seconds to wait for each internal loop iteration

tests/selinux_apply_reboot.yml
```diff
@@ -10,12 +10,8 @@
...
-
-    - name: wait for managed host to come back
-      wait_for_connection:
-        delay: 10
-        timeout: 300
+      reboot:
+        msg: Ansible updates triggered
+        pre_reboot_delay: 2
+        post_reboot_delay: 10
+        reboot_timeout: 300
```
richm, line 17: the `pre_reboot_delay: 2` is for the `sleep 2` in the original - the `post_reboot_delay: 10` is for the `delay: 10` in the original

meta/main.yml
```diff
@@ -10,7 +10,7 @@ galaxy_info:
     - centos
   company: Red Hat, Inc.
   license: GPL-3.0+
-  min_ansible_version: 2.5
+  min_ansible_version: 2.8
```
nhosoi, line 13: Should we bump the min_ansible_version to 2.8 in other roles, as well? The oldest is 2.2... (firewall, postfix and tuned)

nhosoi (approved):
lgtm

meta/main.yml
richm: > Should we bump the min_ansible_version to 2.8 in other roles, as well? The oldest is 2.2... (firewall, postfix and tuned)

I think we'll need to figure out what to do there.  It may be that they still work on old versions of ansible - and there may be some users who are stuck on some old version of ansible and use the role - but we cannot support it

PR #73, opened 2021-03-15
RHELPLAN-69928  Add a note to each module Doc to indicate it is private

nhosoi:
Adding "WARNING: Do not use this module directly! It is only for role internal use." to selogin and selinux_modules_facts module DOCUMENTATION.

Divided into 2 commits to make it easier to create a patch.


richm (approved):
lgtm

PR #76, opened 2021-03-30
Fix issues found by - linters - enable all tests on all repos - remove suppressions

nhosoi:
- Cleaning up yamllint line-length errors
- Emptying unnecessary .ansible-lint skip list

RHELPLAN-73247


tasks/selinux_load_module.yml
```diff
@@ -10,14 +10,19 @@
       copy:
         dest: "{{ tempdir.path }}"
         src: "{{ item.path }}"
+        mode: preserve
 
     - name: Install module with priority
-      command: "semodule -n -s {{ store }} -X {{ priority }} -i {{ tempdir.path }}/{{ item.path | basename }}"
+      command: >-
+        semodule -n -s {{ store }} -X {{ priority }} -i
+          {{ tempdir.path }}/{{ item.path | basename }}
```
bachradsusi, line 18: I understand this is reported by the lint, but it seem to make the code worse for reading/understanding. Could you please explain why it's better to in this case?

tests/tests_all_purge.yml
```diff
@@ -22,7 +22,9 @@
         name: sar-user
 
     - name: Add some mapping
-      shell: echo -e -n "{{ semanage_change }}" | /usr/sbin/semanage -i -
+      shell: |-
+        set -euo pipefail
```
bachradsusi, line 26: Is it possible to `echo` to fail at this point? Or is a general good practice the code needs to follow?

tasks/selinux_load_module.yml
richm: > I understand this is reported by the lint, but it seem to make the code worse for reading/understanding. Could you please explain why it's better to in this case?

The background for this effort is that Galaxy and Automation Hub are becoming quite strict with linting of code imported there in collections.  Rather than just suppressing everything, and therefore suppressing actual problems, we are trying to address all of the issues reported by the linters.

The linter likes yaml lines to be under a certain length.  But we do have some room for adjustment if excessive wrapping makes the code unreadable.  I prefer putting such ignores in-line rather than globally in a .yamllint.yml file.

@nhosoi please put a yamllint line-length ignore here instead of the wrapping.

tests/tests_all_purge.yml
richm: > Is it possible to `echo` to fail at this point? Or is a general good practice the code needs to follow?

It isn't possible for `echo` to fail here unless there is some real garbage data in `semanage_change` - but it is possible for `semanage` to fail.

And yes - part of the "penalty" for using the unrecommended, but sometimes necessary, `shell` module, (in fact ansible-lint complains about this and we have to ignore it - `- '305'  # Use shell only when shell functionality is required`) is that we have to use extra checking for the shell commands - so they have a lint rule for forcing the use of `set -o pipefail` when using `shell` - if we're going to do that, we should add `-e` so that shells exit immediately upon error, and `-u` to catch unset variable usage.  Neither one of those cases apply to the above, but we should be consistent and add both of those everywhere.

bachradsusi: I see, thanks. Would it be possible to update commit messages with these explanations?

richm: > I see, thanks. Would it be possible to update commit messages with these explanations?

It depends.  How detailed would you like these explanations to be?

richm:
There is a warning on line 27 in tasks/selinux_load_module.yml about comment indentation

richm (approved):
lgtm - @bachradsusi ?

nhosoi:
> lgtm - @bachradsusi ?

Thank you, @richm. Please give me one sec. I broke some CI tests in my early commits. I think I fixed them, but I'd like to see all the CI tests finished successfully (hopefully :).

bachradsusi:
lgtm, thanks!

nhosoi:
> > lgtm, thanks!
> ok - if all of the tests pass, I'll merge

Thank you, @bachradsusi! Thank you, @richm!

richm:
The `rhel-x` failure is covered by a bz for RHEL9

PR #77, opened 2021-03-31
tag problematic tests; fix wording; fix formatting

richm:
Some tests are problematic which cause errors in certain cases,
reboot the system, etc.  Tag these tests so the test system can
ignore them in certain cases.

Fix the wording, spelling in a couple of places.

Ensure all tasks are named.

Fix formatting in a couple of places.

### DIFF
--- a/tests/tests_all_purge.yml
+++ b/tests/tests_all_purge.yml
@@ -9,12 +9,14 @@
       login -a -s staff_u sar-user
 
   tasks:
-    - name: Install SELinux tool semanage on Fedora
+    - name: Ensure SELinux tool semanage
       package:
         name:
           - policycoreutils-python-utils
         state: present
-      when: ansible_distribution == "Fedora"
+      when: ansible_distribution == "Fedora" or
+        ( ansible_distribution_major_version | int > 7 and
+          ansible_distribution in ["CentOS", "RedHat"] )
 
     - name: Add a Linux System Roles SELinux User
       user:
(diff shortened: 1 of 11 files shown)


richm:
This supersedes https://github.com/linux-system-roles/selinux/pull/44

richm:
[citest bad]

richm:
[citest pending]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

tests/tests_all_purge.yml
```diff
@@ -9,12 +9,14 @@
       login -a -s staff_u sar-user
 
   tasks:
-    - name: Install SELinux tool semanage on Fedora
+    - name: Ensure SELinux tool semanage
```
nhosoi, line 12: If set_selinux_variables.yml "[ensure SELinux tool semanage](https://github.com/linux-system-roles/selinux/pull/77/files#diff-94524c7699c4ef2c05e1504ca723ff3d38cb0416ff381e7f874a7bc9e6417040R2-R9)" and it's imported at line 37, this `package` is not needed here?

Please note that `package` module is slow even if the package is not downloaded...

richm: The `package` is needed here because of https://github.com/linux-system-roles/selinux/blob/f35251fff0873e536d90e108e5b599703ae2138b/tests/tests_all_purge.yml#L29

> Please note that package module is slow even if the package is not downloaded...

ok - but I would like to get this fix in so we can drop the patch we're carrying downstream

PR #81, opened 2021-04-12
move example playbook to examples/ directory

richm:


spetrosi:
Shouldn't the example naming follow the `example-$FILE-$ROLENAME` convention?

pcahyna:
> Shouldn't the example naming follow the `example-$FILE-$ROLENAME` convention?

The example- part is conveyed by being in the `examples/` directory and `$ROLENAME` is unnecessary, everything inside the selinux repo belongs by definition to the selinux roie.
If the example is going to be installed together with other playbooks from some other roles, one can enforce any convention that is needed in the install script.

richm:
so - ok to merge?

spetrosi (approved):
LGTM, thanks @pcahyna for the clarification

PR #82, opened 2021-05-06
use lazy unmount to fix umount: target is busy

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1945359
EL9 changed the behavior with respect to mounts, so use lazy
unmount when umounting /sys/fs/selinux
It is safe to do this on all platforms.


nhosoi (approved):
Glad to see rhel-x/ansible-2.9 has passed.
lgtm

PR #89, opened 2021-09-14
Bundling community.general.{sefcontext,selogin,seport} modules

nhosoi:
Background: RHEL will support only ansible-core in the future,
in which these selinux modules are not included.

Note: sefcontext.py, selogin.py and seport.py are copied from
community.general to library and black is applied.

### DIFF
--- a/library/sefcontext.py
+++ b/library/sefcontext.py
@@ -1,0 +1,338 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016, Dag Wieers (@dagwieers) <dag@wieers.com>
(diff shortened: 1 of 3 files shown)


bachradsusi (commented):
I guess the motivation for this change is that community.general modules are not available on some platforms. Could this be explained in commit messages so that it's clear why this happened also in future?

Would it be possible to bundle these modules downstream?  e..g Fedora package would require ansible-collection-community-general.noarch while Red Hat Enterprise Linux would bundle these 3 modules?

Using such complicated shell code directly in ansible playbooks seems to be error prone. Would it be possible to move the code to standalone shell wrapper and run the wrapper?

nhosoi:
In the separate discussion, we concluded this change should not be made in the upstream source code. Thus, I'm closing this pr.
Thank you for participating in the discussion, @bachradsusi and @richm.

PR #92, opened 2021-10-02
Fix version comparisons for ansible_distribution_major_version

darxriggs:
Comparisons like `ansible_distribution_major_version > "7"` were broken.
Comparing number strings of different length may result in an unexpected result.
For example `"10" > "7"` evaluates to `false`.
    
Comparisons like `ansible_distribution_major_version | int > 7` were fine.
    
`... is version('7', '>')` could also be used to compare versions.
As the `... | int` conversion is used in many other places, use it for consistency.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -25,7 +25,7 @@
       - policycoreutils-python-utils
     state: present
   when: ansible_distribution == "Fedora" or
-    ( ansible_distribution_major_version > "7" and
+    ( ansible_distribution_major_version | int > 7 and
       ansible_distribution in ["CentOS", "RedHat", "Rocky"] )
 
 - name: Set permanent SELinux state if enabled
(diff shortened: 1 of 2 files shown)


richm:
[citest commit:4e58fa7d7acba94485385811c7e4c3cf33f27b73]

spetrosi:
@darxriggs please rebase with latest master

darxriggs:
Rebase is done.

After having a look at more roles of the collection, I also changed the implementation from using `... is version('7', '>')` to `... | int > 7` for consistency reasons.

richm:
> Rebase is done.
> 
> After having a look at more roles of the collection, I also changed the implementation from using `... is version('7', '>')` to `... | int > 7` for consistency reasons.

yeah - if we were doing version comparisons of versions like `1.215.3a3` then it would make sense to use the `version` test.  But the semantics of `ansible_distribution_major_version` are pretty clear.

richm (approved):
lgtm

PR #71, opened 2021-02-23
CI: Add support for RHEL-9

jharuda:
From now the `rhel-8-y` status is the latest unreleased RHEL-8 and the `rhel-x` status is pre-released RHEL-9.

PR #105, opened 2022-04-13
support gather_facts: false; support setup-snapshot.yml (#71)

richm:
Some users use `gather_facts: false` in their playbooks.  This changes
the role to work in that case, by gathering only the facts it requires
to run.
CI testing can be sped up by creating a snapshot image pre-installed
with packages.  tests/setup-snapshot.yml can be used by a CI system
to do this.


nhosoi (approved):
lgtm

PR #108, opened 2022-05-04
Add seuser and selevel support for selinux_fcontexts

benblasco:
Code and documentation changes to support the RFE #106

PR #120, opened 2022-08-05
add test for fcontext seuser and selevel

richm:


richm:
@bachradsusi @benblasco This is the test for https://github.com/linux-system-roles/selinux/pull/108 - ok?

richm:
[citest]

bachradsusi:
The updated test doesn't test anything from https://github.com/linux-system-roles/selinux/pull/108 and would pass with old role release without support for `seuser` and `selevel`.

There are 2 workflow which need to be tested:

1. test whether a new context mapping uses also SELinux user and level part

after:

    - { target: '/tmp/test_dir1(/.*)?', setype: 'user_home_dir_t', 
        ftype: 'd', seuser: 'staff_u', selevel: 's0-s0:c0.c1023' }

you should get *staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023* instead of *system_u:object_r:user_home_dir_t:s0* in `semanage fcontext` output:

    # semanage fcontext -l -C | grep '/tmp/test_dir1'
    /tmp/test_dir1(/.*)?                               directory          staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023 

2. test whether the relabel forces change of the user part as well

you need to create */tmp/test_dir1* and check the SELinux context before the role is applied, and recheck the context again after the role is applied, see https://bugzilla.redhat.com/show_bug.cgi?id=2115157#c2

benblasco:
There are two new lines of code:
```
    selevel: "{{ item.selevel | default(omit) }}"
    seuser: "{{ item.seuser | default(omit) }}"
```

So from my simplistic point of view there needs to be a test to check the new selevel if the selevel is defined, and a second equivalent test for seuser.

richm:
> The updated test doesn't test anything from #108

I don't understand.  It is testing that `seuser` and `selevel` can be passed to the sefcontext module.

It is testing as much as anything else in tests_fcontext.yml - this is all the checking that the test does: https://github.com/linux-system-roles/selinux/blob/master/tests/tests_fcontext.yml
```
    - name: initial changes
      include_role:
        name: linux-system-roles.selinux
      vars:
        selinux_fcontexts_purge: true
        selinux_fcontexts:
          - { target: '/tmp/test_dir1(/.*)?', setype: 'user_home_dir_t',
              ftype: 'd' }

    - name: include test variables
      import_tasks: set_selinux_variables.yml

    - name: save state after initial changes and before other changes
      set_fact:
        fcontext_before: "{{ selinux_role_fcontext.stdout }}"

    - name: Check if there are SELinux fcontext mapping changes
      assert:
        that: fcontext_before | length > 0
```

AFAICT, this test doesn't actually go out and examine the file system, it just calls `semanage fcontext -l -n -C` to see what the policy is, to see if the policy changed.

> and would pass with old role release without support for `seuser` and `selevel`.

Well sure, because the old role release ignored `seuser` and `selevel` - I'm not sure what point you are making here.

> 
> There are 2 workflow which need to be tested:
> 
>     1. test whether a new context mapping uses also SELinux user and level part
> 
> 
> after:
> 
> ```
> - { target: '/tmp/test_dir1(/.*)?', setype: 'user_home_dir_t', 
>     ftype: 'd', seuser: 'staff_u', selevel: 's0-s0:c0.c1023' }
> ```
> 
> you should get _staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023_ instead of _system_u:object_r:user_home_dir_t:s0_ in `semanage fcontext` output:
> 
> ```
> # semanage fcontext -l -C | grep '/tmp/test_dir1'
> /tmp/test_dir1(/.*)?                               directory          staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023 
> ```
> 
>     2. test whether the relabel forces change of the user part as well
> 
> 
> you need to create _/tmp/test_dir1_ and check the SELinux context before the role is applied, and recheck the context again after the role is applied, see https://bugzilla.redhat.com/show_bug.cgi?id=2115157#c2

So why don't any of the existing tests_fcontext.yml tests do this?  Are the `seuser` and `selevel` parameters somehow different in the way that their application must be checked?

bachradsusi:
> > The updated test doesn't test anything from #108
> 
> I don't understand. It is testing that `seuser` and `selevel` can be passed to the sefcontext module.

So it's supposed to test whether the role still works with *seuser* and *selevel* parameters, I see. I was confused by the commit message.

> Well sure, because the old role release ignored `seuser` and `selevel` - I'm not sure what point you are making here.

This was the point. `seuser` and `selevel` were ignored before and there's no check whether they are actually used or not. The only check is whether there's any change in `semanage fcontext -l -C` - it means that the updated test passes with older role while I expected it would fail as both parameters are ignored.

> 
> So why don't any of the existing tests_fcontext.yml tests do this? Are the `seuser` and `selevel` parameters somehow different in the way that their application must be checked?

Because I'm lazy, writing ansible playbooks is hard for me, and that moment it was sufficient to check whether there's a change in file contexts because  it meant that parameters were applied.

richm:
> > > The updated test doesn't test anything from #108
> > 
> > 
> > I don't understand. It is testing that `seuser` and `selevel` can be passed to the sefcontext module.
> 
> So it's supposed to test whether the role still works with _seuser_ and _selevel_ parameters, I see. I was confused by the commit message.
> 
> > Well sure, because the old role release ignored `seuser` and `selevel` - I'm not sure what point you are making here.
> 
> This was the point. `seuser` and `selevel` were ignored before and there's no check whether they are actually used or not. The only check is whether there's any change in `semanage fcontext -l -C` - it means that the updated test passes with older role while I expected it would fail as both parameters are ignored.
> 
> > So why don't any of the existing tests_fcontext.yml tests do this? Are the `seuser` and `selevel` parameters somehow different in the way that their application must be checked?
> 
> Because I'm lazy, writing ansible playbooks is hard for me, and that moment it was sufficient to check whether there's a change in file contexts because it meant that parameters were applied.

Then the test I have in this PR is valid in the context of the current tests.

and perhaps we should revisit the selinux role tests in subsequent PRs to actually grab the context from the object being managed (e.g. use ls -Z to examine file/directory contexts; use ps -Z to examine port contexts; etc.) to compare with the expected context.

bachradsusi:
Could you please add this or similar check:

~~~
--- a/tests/tests_fcontext.yml
+++ b/tests/tests_fcontext.yml
@@ -24,6 +24,10 @@
       assert:
         that: fcontext_before | length > 0
 
+    - name: Check if there's mapping to staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023
+      assert:
+        that: '"staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023" in fcontext_before'
+
~~~

richm:
[citest]

richm:
> Could you please add this or similar check:
> 
> ```
> --- a/tests/tests_fcontext.yml
> +++ b/tests/tests_fcontext.yml
> @@ -24,6 +24,10 @@
>        assert:
>          that: fcontext_before | length > 0
>  
> +    - name: Check if there's mapping to staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023
> +      assert:
> +        that: '"staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023" in fcontext_before'
> +
> ```

how's this?  https://github.com/linux-system-roles/selinux/pull/120/commits/48e252267217bc658e30e1368134abcc7b4b8ada

ukulekek:
[citest]

tests/tests_fcontext.yml
```diff
@@ -1,17 +1,32 @@
...
+      /tmp/test_dir2[^ ]+[ ]+directory[
+      ]+staff_u:object_r:user_home_dir_t:s0-s0:c0.c1023
   tasks:
 
+    - name: include test variables
+      import_tasks: set_selinux_variables.yml
+
+    - name: Ensure no mapping to seuser/selevel context
+      assert:
+        that: not __str_before in selinux_role_fcontext.stdout
```
bachradsusi, line 20: Given that there's `selinux_fcontexts_purge: true` bellow, I don't think it's necessary as all local fcontext modification should be dropped by the *initial changes* task

bachradsusi:
Works for me, thanks!

tests/tests_fcontext.yml
richm: dropped check before initial changes purge

PR #123, opened 2022-09-06
 WIP: add 'local' parameter to seport

richm:
- `local: true`:
  * `state: present` enforces change to be made even though the
  port mapping could already exists in built in policy
  *`state: absent` would remove only local modification and would not
  try to remove builtin mapping.

Implemented using local_seport module which is copied from seport module
and update to accept `local`


richm:
[citest]

richm:
closing in favor of https://github.com/linux-system-roles/selinux/pull/124

Issue #133, opened 2023-01-09
Load module task not idempotent


jamieburchell:
Example config:

```
selinux_modules:
  - path: "selinux_modules/php_ptrace.pp"
    state: enabled
```

The task makes the same changes each time it runs, regardless of whether the module is already installed:

```
TASK [linux-system-roles.selinux : Create temporary directory] ***************************************************************************************************************************
changed: [example_01]

TASK [linux-system-roles.selinux : Copy module file "selinux_modules/php_ptrace.pp"] *****************************************************************************************************
changed: [example_01]

TASK [linux-system-roles.selinux : Install module with priority] *************************************************************************************************************************
skipping: [example_01]

TASK [linux-system-roles.selinux : Install module] ***************************************************************************************************************************************
changed: [example_01]

TASK [linux-system-roles.selinux : Clean temporary directory "/tmp/linux_system_role.selinux4smkzw61"] ***********************************************************************************
changed: [example_01]
```

bachradsusi:
Since SELinux userspace release 3.4 or since RHEL 8.6, it's possible to list also installed module checksums. In case that both installed module and local file checksums are same, the role could not install module again, e.g. for zosremote.pp

~~~
# file_checksum="sha256:`/usr/libexec/selinux/hll/pp zosremote.pp | sha256sum | awk '{ print \$1;}'`"
# echo $file_checksum 
sha256:77d9792d5964f3046a6c3e24c105b73a51de67b7a0cda79783b9f175abec21e9

# module_checksum=`semodule -l --checksum | grep zosremote | awk '{ print $2; }'`
# echo $module_checksum 
sha256:77d9792d5964f3046a6c3e24c105b73a51de67b7a0cda79783b9f175abec21e9

# if [ "$module_checksum" = "$file_checksum" ]; then echo "already installed"; else semodule -i zosremote.pp; fi
already installed
~~~

jamieburchell:
I've seen other roles handle this by keeping the source te/pp files on the remote location and only proceeding to install when the files have changed.

bachradsusi:
The source files on the remote location would not be reliable. Between two role runs, users could use `semanage` to manage modules directly on the remote location .

richm:
> > The source files on the remote location would not be reliable. Between two role runs, users could use `semanage` to manage modules directly on the remote location .
> 
> Should users be doing that if the modules are Ansible managed?

[...]

PR #135, opened 2023-01-13
Rewrite selinux_load_module.yml to use local_semodule 

bachradsusi:
the new implementation compares checksum of files which are supposed to be installed with checksum of already installed modules. If they match it's not necessary to install the file again.

For more details see https://github.com/linux-system-roles/selinux/issues/133

### DIFF
--- a/library/local_semodule.py
+++ b/library/local_semodule.py
@@ -1,0 +1,402 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023, Petr Lautrbach <lautrbach@redhat.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: local_semodule
+short_description: Manages SELinux modules
+description:
+    - Manages SELinux modules
+options:
+  name:
+    description:
+      - A module name, used with "enabled" to enabled disabled module
+    type: str
+  path:
+    description:
+      - A local module filename, used with "enabled" to install a module
+    type: str
+  priority:
+    description:
+      - SELinux module priority, 0 means SELinux store default (400) with "path" and "enabled", otherwise all priorities
+    type: int
+    default: 0
+  state:
+    description:
+      - Desired module state
+      - enabled - install a module file with `path`, enable a module with `name`
+      - disabled - disable module
+      - absent - remove module
+    type: str
+    choices: [ enabled, disabled, absent ]
+    default: enabled
+  store:
+    description:
+      - Use other than current SELinux module store, e.g. mls, minimum, refpolicy ...
+    type: str
+notes:
+   - The changes are persistent across reboots.
+   - Not tested on any debian based system.
+requirements:
+- python3-libselinux
+- python3-libsemanage
+author:
+- Petr Lautrbach (@bachradsusi)
+"""
+
+EXAMPLES = r"""
+- name: Install mymodule.pp with default priority (400)
+  local_semodule:
+    path: mymodule.pp
+    state: enabled
+
+- name: Install mymodule.pp with default priority (300)
+  local_semodule:
+    path: mymodule.pp
+    priority: 300
+    state: enabled
+
+- name: Disable mymodule
+  local_semodule:
+    name: mymodule
+    state: disabled
+
+- name: Enable mymodule
+  local_semodule:
+    name: mymodule
+    state: enabled
+
+- name: Remove mymodule with priority 300
+  local_semodule:
+    name: mymodule
+    priority: 300
+    state: absent
+
+- name: Remove mymodule - all priorities
+  local_semodule:
+    name: mymodule
+    state: absent
+"""
+
+import traceback
+
+SELINUX_IMP_ERR = None
+try:
+    import selinux
+
+    HAVE_SELINUX = True
+except ImportError:
+    SELINUX_IMP_ERR = traceback.format_exc()
+    HAVE_SELINUX = False
+
+SEMANAGE_IMP_ERR = None
+try:
+    import semanage
+
+    HAVE_SEMANAGE = True
+except ImportError:
+    SEMANAGE_IMP_ERR = traceback.format_exc()
+    HAVE_SEMANAGE = False
+
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+
+import os.path
+
+
+def get_runtime_status(ignore_selinux_state=False):
+    return ignore_selinux_state or selinux.is_selinux_enabled()
+
+
+def init_libsemanage(store=""):
+    sh = semanage.semanage_handle_create()
+    if not sh:
+        raise ValueError("Could not create semanage handle")
+
+    if store != "":
+        semanage.semanage_select_store(sh, store, semanage.SEMANAGE_CON_DIRECT)
+    semanage.semanage_connect(sh)
+    semanage.semanage_set_reload(sh, 0)
+    return sh
+
+
+def semodule_install(module, path, priority, sh):
+    """Install a local policy module
+
+    :type module: AnsibleModule
+    :param module: Ansible module
+
+    :type path: str
+    :param name: a local module file (either .cil or .pp) to be installed on a node
+
+    :type priority: int
+    :param priority: SELinux module priority, 0 - use SELinux module store default (400)
+    """
+
+    path_checksum = "sha256:" + module.sha256(path)
+
+    filename = os.path.split(path)[1]
+    (name, ext) = os.path.splitext(filename)
+
+    if ext == ".cil":
+        cil = 1
+    elif ext == ".pp":
+        cil = 0
+    else:
+        raise Exception("Only .cil and .pp files are supported")
+
+    result = {
+        "changed": False,
+        "name": name,
+        "priority": priority,
+        "state": "enabled",
+    }
+
+    try:
+        r, modkey = semanage.semanage_module_key_create(sh)
+    except AttributeError:
+        # fallback to old libsemanage api
+        modkey = None
+
+    if modkey is not None and r != 0:
+        raise Exception(r)
+
+    if modkey:
+        r = semanage.semanage_module_key_set_name(sh, modkey, name)
+        if r != 0:
+            raise Exception(r)
+
+        if priority == 0:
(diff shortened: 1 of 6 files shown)


tasks/selinux_load_module.yml
```diff
@@ -14,57 +14,40 @@
...
-    # yamllint disable rule:line-length
-    - name: Install module with priority
-      command: >-
-        semodule -n -s {{ store }} -X {{ priority }} -i {{ tempdir.path }}/{{ item.path | basename }}
-      when: priority != "0" and selinux_priorities
+    - name: Install the module
+      local_semodule:
+        path: "{{ tempdir.path }}/{{ item.path | basename }}"
+        priority: "{{ priority }}"
+        state: "enabled"
```
richm, line 21:
```suggestion
        state: enabled
```
https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#yaml-and-jinja2-syntax
```
Do not use quotes unless you have to, especially for short module-keyword-like strings like present, absent, etc. But do use quotes for user-side strings such as descriptions, names, and messages.
```
I realize there are other places in the selinux yaml files where quotes are used, so perhaps not using quotes in this case makes this not consistent with other places in the file, but it is preferred not to use quotes in this case for new code.

tasks/selinux_load_module.yml
```diff
@@ -14,57 +14,40 @@
...
-  notify: __selinux_reload_policy
-
-- name: Remove module with priority
-  command: "semodule -n -s {{ store }} -X {{ priority }} -r {{ item.name }}"
-  when: state == "absent" and current_state != "absent" and
-        (priority != "0" and selinux_priorities)
+- name: Remove module
+  local_semodule:
+    name: "{{ item.name }}"
+    state: "absent"
```
richm, line 32:
```suggestion
    state: absent
```

tasks/selinux_load_module.yml
```diff
@@ -14,57 +14,40 @@
...
+    priority: "{{ priority }}"
+  when: state == "absent" and current_state != "absent"
   notify: __selinux_reload_policy
 
 - name: Enable module
-  command: "semodule -n -s {{ store }} -e {{ item.name }}"
+  local_semodule:
+    name: "{{ item.name }}"
+    state: "enabled"
```
richm, line 40:
```suggestion
    state: enabled
```

tasks/selinux_load_module.yml
```diff
@@ -14,57 +14,40 @@
...
-  command: "semodule -n -s {{ store }} -d {{ item.name }}"
-  when: state == "disabled" and (priority == "0" or not selinux_priorities)
-  notify: __selinux_reload_policy
-
-- name: Remove module
-  command: "semodule -n -s {{ store }} -r {{ item.name }}"
-  when: state == "absent" and (priority == "0" or not selinux_priorities)
+  local_semodule:
+    name: "{{ item.name }}"
+    state: "disabled"
```
richm, line 48:
```suggestion
    state: disabled
```

richm:
Can you add some sort of test for this?

bachradsusi:
> Can you add some sort of test for this?

Is there a way to check whether a handler was run?

~~~
RUNNING HANDLER [linux-system-roles.selinux : Reload SELinux policy] ...
~~~

If this happen after the role is run second time with the same module, the change didn't work (or it's RHEL < 9)

bachradsusi:
> > Can you add some sort of test for this?
> 
> Is there a way to check whether a handler was run?
> 
> ```
> RUNNING HANDLER [linux-system-roles.selinux : Reload SELinux policy] ...
> ```
> 
> If this happen after the role is run second time with the same module, the change didn't work (or it's RHEL < 9)

Probably not necessary. I'll use `/var/lib/selinux/active/commit_num` which is raised every time a policy is changed/rebuilt

richm:
> > > Can you add some sort of test for this?
> > 
> > 
> > Is there a way to check whether a handler was run?
> > ```
> > RUNNING HANDLER [linux-system-roles.selinux : Reload SELinux policy] ...
> > ```
> >   
> > If this happen after the role is run second time with the same module, the change didn't work (or it's RHEL < 9)
> 
> Probably not necessary. I'll use `/var/lib/selinux/active/commit_num` which is raised every time a policy is changed/rebuilt

It's very tricky to try to figure out if the handler has been called.  Handlers operate in a different execution context than the main flow of the playbook.  You might be able to set some sort of global variable in the handler and see if that variable is set, but even then there are issues with variable scopes.

bachradsusi:
Quotes fixed, test added. PTAL

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,56 @@
+---
+- name: Test management of SELinux modules
+  hosts: all
+  become: true
+  become_method: sudo
```
richm, line 5:
```suggestion
```

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,56 @@
...
+  become_user: root
```
richm, line 6:
```suggestion
```

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,56 @@
...
+  gather_facts: true
```
richm, line 7:
```suggestion
```

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,56 @@
...
+  tasks:
+    - name: Execute the role and catch errors
+      when: ansible_distribution == "Fedora" or
+        (ansible_distribution_major_version | int > 8 and
+        ansible_distribution in ["CentOS", "RedHat", "Rocky"]) or
+        (ansible_distribution_major_version | int == 8 and
```
richm, line 13: did you mean `int <= 8` here?

For this `when` condition, do you want to execute on Fedora, or on Red Hat family OS with version in [6, 7, 8, 9]?

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,56 @@
...
+          shell: cat /var/lib/selinux/targeted/active/commit_num
+          register: commit_num_before
+        - name: Execute the role
+          include_role:
+            name: linux-system-roles.selinux
+          register: role_result
+        - name: Get current commit_num
+          shell: cat /var/lib/selinux/targeted/active/commit_num
+          register: commit_num_after
+      rescue:
```
richm, line 34: I don't think you need this `rescue` block - if one of the tasks in the `block` list fails, it will report an error, and the `always` block tasks will execute.

library/local_semodule.py
```diff
@@ -0,0 +1,367 @@
...
+        raise Exception(r)
+
+    r = semanage.semanage_module_key_set_name(sh, modkey, name)
+    if r != 0:
+        raise Exception(r)
+
+    try:
+        r, m_enabled = semanage.semanage_module_get_enabled(sh, modkey)
+    except Exception:
+        raise ValueError("Could not find module {name}".format(name=name))
```
richm, line 240: Is there any information in the exception, like an error message, that might be helpful to display to the user, to try to fix the problem?

richm:
[citest]

bachradsusi:
[citest]

bachradsusi:
[citest]

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,41 @@
...
+- name: Test management of SELinux modules
+  hosts: all
+  become: true
+  tasks:
+    - name: Execute the role and catch errors
+      when: ansible_distribution == "Fedora" or
+        (ansible_distribution_major_version | int > 8 and
+        ansible_distribution in ["CentOS", "RedHat", "Rocky"]) or
+        (ansible_distribution_major_version | int == 8 and
+        ansible_distribution_version >= "8.6" and
```
richm, line 11:
```suggestion
        ansible_distribution_version is version("8.6", ">=") and
```

bachradsusi:
[citest]

tests/tests_selinux_modules_checksum.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Test management of SELinux modules
+  hosts: all
+  become: true
+  tasks:
+    - name: Execute the role and catch errors
+      when: ansible_distribution == "Fedora" or
+        (ansible_distribution in ["CentOS", "RedHat", "Rocky"] and
+        ansible_distribution_version is version("8.6", ">="))
```
richm, line 9: This new `local_semodule` Ansible module will only support RHEL 8.6 and later?  Is that because of the `selinux` and `semanage` python library dependencies?  Will users of RHEL 6, RHEL 7, and RHEL 8 still be able to manage their selinux modules with this role?

bachradsusi:
[citest]

bachradsusi:
[citest]

bachradsusi:
I wasn't able to reproduce  tests_all_transitions FAILED problems on Fedora 37 or CentOS 7. I think it's not related to this PR. All other tests passed

tests/tests_selinux_modules_checksum.yml
richm: @bachradsusi ^^^ ?

bachradsusi:
My last two comments are pending and it looks like they are not visible?

2 days ago I wrote:

This new test covers only the issue which triggered this change - https://github.com/linux-system-roles/selinux/issues/133 The load module task will be idempotent on rhel-8.6 and later due to the fact that libsemanage used on older systems does not support checksums. And it would be too complicated to implement it in the module for these old systems.

But the module and the role are still able to manage modules on RHEL 6, RHEL 7, and RHEL 8 - this is covered by tests_selinux_modules.yml


A while ago I wrote:

I'm not sure what is the question.

local_semodule Ansible module supports all rhels since rhel6. But only since rhel-8.6, the load module task is idempotent as older versions of libsemanage does not support checksums. And I don't think it's worth to implement checksum functionality directly in local_semodule for these old systems.

tests_selinux_modules_checksum.yml tests only https://github.com/linux-system-roles/selinux/issues/133 on systems rhel-8.6 and later. Other module operations are covered by tests_selinux_modules.yml

bachradsusi:
oh, it was something with review, and I had to open changes and Submit review to get comments visible.

tests/tests_selinux_modules_checksum.yml
richm: ok - then we need to add something to the README.md https://github.com/linux-system-roles/selinux#manage-selinux-modules - something like
"NOTE: Managing modules is idempotent only on Fedora, and EL 8.6 and later.  You can manage modules on old releases, but it will not be idempotent."

richm (approved):

PR #139, opened 2023-01-24
ensure semanage present on EL7 tests; fix jinja spacing, quoting

richm:
Ensure that the semanage command is present on EL7 for tests.
Fix some Jinja spacing, quoting issues


richm:
[citest]

PR #145, opened 2023-01-31
do not use 'become' in tests, examples

richm:
The main reason is that the baseos ci test runner does not have `sudo`
installed by default, so we would need to use a different `become`
method.
But since we have to modify all of these anyway, best to just get
rid of `become`.  In general, you can't really use this role without
being root on the managed node, so it is implied.  Ansible recommends
to let the user specify `become` if necessary.  There are a variety
of ways the user can specify become:

* ansible-playbook CLI `--become` et. al.

https://docs.ansible.com/ansible/latest/cli/ansible-playbook.html

* connection variables

https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#connection-variables

* environment variables, configuration settings

https://docs.ansible.com/ansible/latest/reference_appendices/config.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ vars:
   [ see below ]
 roles:
   - role: linux-system-roles.selinux
-    become: true
 ```
 
 #### purge local modifications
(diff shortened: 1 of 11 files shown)


richm:
[citest]

Issue #142, opened 2023-01-26
Make all tasks idempotent


richm:

bachradsusi:
I think it could work without a permanent storage. But I'm not Ansible expert and currently I have no idea how to implement it:

1. get module file sha256 on the host, currently it's counted in `local_semodule.py` on a node, but it can be simply counted on the host using `sha256sum` or pythons `hashlib`
2. get installed module checksum on  the node like in the `local_semodule.py` - https://github.com/linux-system-roles/selinux/blob/main/library/local_semodule.py#L192 , or using `semodule -lfull --checksum`. I guess this could be also included into facts generated by `selinux_module_facts.py`
3. compare values and if they differ, copy the module to the node and install

richm:
There are hashing filters built-in to Ansible: https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters.html#hashing-filters
So the role could use the `file` lookup to read the contents: https://docs.ansible.com/ansible/2.9/plugins/lookup/file.html
Then use one of the hashing filters to generate the hash

bachradsusi:
I've started with https://github.com/bachradsusi/linux-system-roles-selinux/tree/selinux_facts

With this change, ansible_installed_modules fact would contain also SELinux module checksums which could be compared with a value generated by the hashing function and if they're not same, the role could install the file.

richm:
> I've started with https://github.com/bachradsusi/linux-system-roles-selinux/tree/selinux_facts
> 
> With this change, ansible_installed_modules fact would contain also SELinux module checksums which could be compared with a value generated by the hashing function and if they're not same, the role could install the file.

Looks good - this will work on EL6, 7, 8, and 9?  I'm assuming the checksum field will be blank on versions less than 8.6?

bachradsusi:
yes. checksums will be blank on rhel8.6 and older. also priorities will be 0 on old systems which doesn't support it - rhel-6 and early versions of rhel7. the only visible change will be different structure of selinux module facts. instead of ` [name][priority] = "enabled"` it will be `[name][priority] = { enabled: True, checksum: "12345" }`

jamieburchell:
> @jamieburchell this is fixed - will soon be publishing the selinux role 1.5.3 to Galaxy - can you give it a try?

I'm getting:

```
TASK [linux-system-roles.selinux : Get checksum for selinux_modules/php_ptrace.pp] **************************************************************************************
[...]

PR #148, opened 2023-02-02
Use lookup file + sha256 to get hash of local policy file

richm:
This is a follow up to https://github.com/linux-system-roles/selinux/issues/142
The `delegate_to: localhost`, if using `become` with the playbook, is causing
a `sudo` to `localhost`.  Use a `file` lookup through a `hash` filter to
get the checksum instead and try to bypass that delegation.


richm:
[citest]

richm:
[citest]

richm:
[citest]

PR #152, opened 2023-02-08
Use stat on localhost with become: false

richm:
The `file` lookup does not work on EL7 with binary files.
Instead, use `stat` delegated to `localhost` with `become: false`
which should use the userid of the user running Ansible to
read the file, much like the `file` lookup.


richm:
[citest]

PR #155, opened 2023-02-09
use fileglob to lookup selinux module file

richm:
use `fileglob` to lookup selinux module file.  This should use the
same lookup path that `copy` `src` uses so that we can get the
same semantics.  The wrinkle is that the filename must have file
glob characters escaped so that we do not do an actual `glob`
lookup.

Some of the selinux module files used in testing are moved to the
`tests/files/selinux_modules/` directory so that we can test
the file lookups, and some of the tests use an absolute path so
we can verify that works too.  One file was left in place to
ensure that that scenario continues to work.

A test was added to ensure that if a file name is passed that
uses fileglob characters, it will be escaped and not matched.


richm:
[citest]

jamieburchell:
This seems to work fine in my environment 👍

richm:
> This seems to work fine in my environment +1

Thanks for checking!  And thanks for your help and patience with this issue.

PR #163, opened 2023-04-07
ansible-lint - use changed_when for conditional tasks

richm:
ansible-lint now requires the use of changed_when even for
conditional tasks


richm:
[citest]

PR #173, opened 2023-05-22
ci: Fix commitlint to read special characters '' "" ** in PR titles

spetrosi:


.github/workflows/commitlint.yml
```diff
@@ -39,6 +39,8 @@ jobs:
           --to '${{ github.event.pull_request.head.sha }}' --verbose
 
       - name: Run commitlint on PR title
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.title }}
         run: >-
-          echo '${{ github.event.pull_request.title }}' |
-          npx commitlint --verbose
+          echo "$PR_NUMBER" | npx commitlint --verbose
```
richm, line 45: So why doesn't `echo '${{ github.event.pull_request.title }}'` work?  It should put all of the special characters in single quotes in the shell, which should be fine

bachradsusi: Does it mean that PR's tittle has to follow the same rules as commits?

bachradsusi: Why? Is it documented anywhere?

spetrosi: > So why doesn't echo '${{ github.event.pull_request.title }}' work? It should put all of the special characters in single quotes in the shell, which should be fine

When there is a single quote in `{{ github.event.pull_request.title }}` it closes the opening quote.

spetrosi: > Does it mean that PR's tittle has to follow the same rules as commits?

Yes, this is required because we collect PR titles and descriptions into release notes. Because PRs might have multiple commits, it's better to collect PR title and description, not each commit.
This also helps consistency and repo hygiene in GitHub.

> Why? Is it documented anywhere?

Not yet, it should be documented in https://linux-system-roles.github.io/contribute.html#conventional-commits-format mentions now after your comment I understood that the doc says "We use an automation that collects commit messages into the changelog and release notes", while it should instead say that we collect PR title and description to commit messages.

spetrosi: Adding info about PR titles and descriptions to the contribute doc in https://github.com/linux-system-roles/linux-system-roles.github.io/pull/94

spetrosi: > > So why doesn't echo '${{ github.event.pull_request.title }}' work? It should put all of the special characters in single quotes in the shell, which should be fine
> 
> When there is a single quote in `{{ github.event.pull_request.title }}` it closes the opening quote.

Oh wait, you are right, everything works as is. I confused to something else. Verified on this PR. False alarm then, closing.

bachradsusi: Thanks for the explanation and the documentation update.

PR #184, opened 2023-08-14
docs: Add example playbook and "readme" for confined users

vmojzis:
Note: Relabeling user home dirs should probably be added to https://docs.ansible.com/ansible/latest/collections/community/general/selogin_module.html

### DIFF
--- a/README-confined_users.md
+++ b/README-confined_users.md
@@ -1,0 +1,61 @@
+# Confined users
+
+By default linux users are mapped to SELinux user `unconfined_u`, which is subject to minimal restrictions. You can greatly improve security of your systems by confining users, that is by mapping them to SELinux users with less privilege.
+`confined_users-playbook.yml` contains an example configuration for confining existing users and creating new ones with varying levels of access.
+
+## Expected functionality
+
+* Set default user mapping
+* Confine existing users accounts
+* Set booleans to customize access
+
+## Variables
+
+### selinux_booleans
+
+SELinux policy provides several booleans for customizing access of confined users.
+
+`selinux_booleans` is expected to hold a list of booleans together with their intended value and persistence setting.
+Formally it is a `list` of `dict`, where each `dict` is in the same format as used by the
+[seboolean](https://docs.ansible.com/ansible/latest/collections/ansible/posix/seboolean_module.html#ansible-collections-ansible-posix-seboolean-module)
+module.
+
+```yaml
+selinux_booleans:
+  - {name: 'ssh_sysadm_login', state: 'on', persistent: 'yes'}
+  - {name: 'user_exec_content', state: 'off', persistent: 'yes'}
+```
+
+See
+[confined users documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/using_selinux/index#managing-confined-and-unconfined-users_using-selinux)
+ for a list of relevant booleans.
+
+### existing_privileged_users
+
+Specify a list of existing users to be mapped to `staff_u` and allowed to use `sudo` to perform administrative tasks.
+
+```yaml
+existing_privileged_users:
+  - "Mark"
+  - "Roger"
+```
+
+### selinux_logins
+
+Manage the linux user to SELinux user mapping. This is a `list` of `dict`,
+where each `dict` is in the same format as used by the
+[selogin](https://docs.ansible.com/ansible/latest/collections/community/general/selogin_module.html)
+module.
+
+```yaml
+selinux_logins:
+  - {login: 'plautrba', seuser: 'staff_u', state: 'absent'}
+  - {login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023', state: 'present'}
+```
+
+`__default__` is the value assigned to a new user when no login mapping is specified for them. All other lines must correspond to existing user accounts. Note that changing a login mapping changes the file context definitions of the user's home directory. This change is applied using `restorecon` at the end of the playbook.
+
+## Examples
+
+Confining users is demonstrated in
+[confined_users-playbook.yml](examples/confined_users-playbook.yml) playbook.
--- a/examples/confined_users-playbook.yml
+++ b/examples/confined_users-playbook.yml
@@ -1,0 +1,126 @@
+---
+- name: Manage SELinux policy example
+  hosts: all
+  become: true
+  vars:
+    # Use "targeted" SELinux policy type
+    selinux_policy: targeted
+    # Set "enforcing" mode
+    selinux_state: enforcing
+    # Set SELinux booleans
+    selinux_booleans:
+      # Allow ssh logins as sysadm_r:sysadm_t
+      # - {name: 'ssh_sysadm_login', state: 'on', persistent: 'yes'}
+      # Allow sysadm_u/staff_u to execute applications in their
+      # home directories and /tmp
+      - {name: 'sysadm_exec_content', state: 'on', persistent: 'yes'}
+      - {name: 'staff_exec_content', state: 'on', persistent: 'yes'}
+      # But block normal users from the same
+      - {name: 'user_exec_content', state: 'off', persistent: 'yes'}
+      # Allow staff user to create and transition to svirt domains
+      # - {name: 'staff_use_svirt', state: 'on', persistent: 'yes'}
+      # Allow confined users the ability to execute the ping and traceroute
+      # - {name: 'selinuxuser_ping', state: 'on', persistent: 'yes'}
+      # Allow users to connect to the local mysql server
+      # - {name: 'selinuxuser_mysql_connect_enabled', state: 'on',
+      #    persistent: 'yes'}
+      # Allow samba and winbind-rpcd to share users home directories
+      # - {name: 'samba_enable_home_dirs', state: 'on', persistent: 'yes'}
+      # Disallow programs, such as newrole, from transitioning to administrative
+      # user domains
+      # - {name: 'secure_mode', state: 'on', persistent: 'yes'}
+
+    # List of existing users to be mapped to staff_u and allowed to use sudo
+    existing_privileged_users:
+      - "Lizzy"
+      - "Tarja"
+
+    # The following section specifies mapping of EXISTING system users
+    # to SELinux users
+    selinux_logins:
+      # Map all new users to SElinux user "user_u" (instead of "unconfined_u")
+      - {login: '__default__', seuser: 'user_u', serange: 's0',
+         state: 'present'}
+      # Alternatively map all new users to SElinux user "staff_u" (instead of
+      # "unconfined_u")
+      # - {login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023',
+      #    state: 'present'}
+      # Map admin users to SElinux user "staff_u"
+      # Map select non-admin privileged users to SELinux user "staff_u"
+      # - {login: '<existing_user>', seuser: 'staff_u', serange: 's0',
+      #    state: 'present'}
+      # - {login: 'Lizzy', seuser: 'staff_u', serange: 's0',
+      #    state: 'present'}
+
+  # Prepare the prerequisites required for this playbook
+  tasks:
+    # Enable kiosk mode -- add extremely limited users "guest" and "xguest"
+    # - name: Install xguest package
+    #  yum:
+    #    name: xguest
+    #    state: present
+
+    # Create login mappings for "existing_privileged_users"
+    - name: Build list of selinux logins
+      set_fact:
+        selinux_logins: "{{ selinux_logins + login_helper | list }}"
+      with_items: "{{ existing_privileged_users }}"
+      vars:
+        login_helper:
+          - {login: "{{ item }}", seuser: 'staff_u', serange: 's0',
+             state: 'present'}
+
+    # Users that get assigned a new SELinux user need to have their homedirs
+    # relabeled.
+    # Gather homedirs of each user with a new login mapping
+    - name: Gather homedirs of users with new login mapping
+      shell: >-
+        # exit in case of errors in the following command ("-o pipefail"
+        # cannot be used since "getent passwd __default__" is expected to fail)
+        set -eu;
+        getent passwd "{{ item.login }}" | awk -F: '{ print $6 }'
+      register: __login_homedirs
+      with_items: "{{ selinux_logins }}"
+      changed_when: false
+
+    # Relabel the gathered homedirs using "selinux_restore_dirs" variable
+    - name: Relabel homedirs of users with new login mapping
+      set_fact:
+        selinux_restore_dirs: "{{ selinux_restore_dirs | default([])
+                                  + [item.stdout] }}"
+      when: item.stdout != ""
+      with_items: "{{ __login_homedirs.results if __login_homedirs.results is
+                   iterable else [] }}"
+
+    # Allow existing_privileged_users to use sudo
+    - name: Allow privileged users to use sudo
+      copy:
+        dest: /etc/sudoers.d/{{ item }}
+        content: "{{ item }} ALL=(ALL) TYPE=sysadm_t ROLE=sysadm_r ALL"
+        mode: "0640"
+      loop: "{{ existing_privileged_users | list }}"
+
+    # Run selinux role to apply changes described in "vars" section
+    - name: Execute the role and catch errors
+      block:
+        - name: Include selinux role
+          include_role:
+            name: linux-system-roles.selinux
+      rescue:
+        # Fail if failed for a different reason than selinux_reboot_required.
+        - name: Handle errors
+          fail:
+            msg: "role failed"
+          when: not selinux_reboot_required
+
+        - name: Restart managed host
+          reboot:
+
+        - name: Wait for managed host to come back
+          wait_for_connection:
+            delay: 10
+            timeout: 300
+
+        - name: Reapply the role
+          include_role:
+            name: linux-system-roles.selinux
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -84,20 +84,6 @@
     ignore_selinux_state: "{{ ansible_selinux.status == 'disabled' }}"
   with_items: "{{ selinux_fcontexts }}"
 
-- name: Restore SELinux labels on filesystem tree
-  command: /sbin/restorecon -R -F -v {{ restorecon_threads }} {{ item }}
-  with_items: "{{ selinux_restore_dirs }}"
-  register: restorecon_cmd
-  changed_when: '"Relabeled" in restorecon_cmd.stdout'
-
-- name: Restore SELinux labels on filesystem tree in check mode
-  command: /sbin/restorecon -R -F -v -n {{ restorecon_threads }} {{ item }}
-  with_items: "{{ selinux_restore_dirs }}"
-  register: restorecon_cmd
-  changed_when: '"Would relabel" in restorecon_cmd.stdout'
-  check_mode: false
-  when: ansible_check_mode
-
 - name: Set an SELinux label on a port
   local_seport:
     ports: "{{ item.ports }}"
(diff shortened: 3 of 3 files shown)


richm:
Just a few minor ansible-lint issues, otherwise, lgtm
would like @bachradsusi to review for correctness

spetrosi:
Why not include this into the main README.md?

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,107 @@
...
+    #    name: xguest
+    #    state: present
+
+    # Add new system users that will be assigned to staff_u SELinux user
+    - name: Add non-root privileged users
+      user:
+        comment: Linux System Roles SELinux User
+        name: "{{ item }}"
+        seuser: staff_u
+        group: "wheel"
```
bachradsusi, line 62: I'd say that `wheel` group is not necessary given that all privileged users have their own sudo config line.

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,107 @@
...
+      #- {name: 'selinuxuser_mysql_connect_enabled', state: 'on', persistent: 'yes'}
+      # Allow samba and winbind-rpcd to share users home directories
+      #- {name: 'samba_enable_home_dirs', state: 'on', persistent: 'yes'}
+      # Disallow programs, such as newrole, from transitioning to administrative user domains
+      #- {name: 'secure_mode', state: 'on', persistent: 'yes'}
+
+    # List of privileged users to be created
+    privileged_users:
+      - "staff"
+      - "roger"
```
bachradsusi, line 34: README uses "mark" and "roger"

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    # The following section specifies mapping of EXISTING system users to SELinux users
+    selinux_logins:
+      # Map all new users to SElinux user "user_u" (instead of "unconfined_u")
+      - {login: '__default__', seuser: 'user_u', serange: 's0', state: 'present'}
+      # Alternatively map all new users to SElinux user "staff_u" (instead of "unconfined_u")
+      #- {login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023', state: 'present'}
+      # Map admin users to SElinux user "sysadm_u"
+      - {login: 'root', seuser: 'sysadm_u', serange: 's0-s0:c0.c1023',
+               state: 'present'}
```
bachradsusi, line 44: In "mls" we use "root" SELinux user for "root" user and `sysadm_u` for admin non-root users. I'd follow it here as well.

examples/confined_users-playbook.yml
bachradsusi: I'd drop creating users completely, expect that users are already created/available and use the role to create new login mappings - https://github.com/linux-system-roles/selinux/blob/main/README.md#selinux_logins

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,107 @@
...
+    - name: Gather homedirs of users with new login mapping
+      shell: >
+        getent passwd "{{ item.login }}" | awk -F: '{ print $6 }'
+      register: __login_homedirs
+      with_items: "{{ selinux_logins }}"
+    # Relabel the gathered homedirs (-F is necessary to change the user portion of file labels)
+    - name: Relabel homedirs of users with new login mapping
+      shell: "restorecon -RF {{ item.stdout }}"
+      when: item.stdout != ""
+      with_items: "{{ __login_homedirs.results if __login_homedirs.results is iterable else [] }}"
```
bachradsusi, line 107: Would it be possible to use https://github.com/linux-system-roles/selinux/blob/main/README.md#selinux_restore_dirs instead of `restorecon` command?

vmojzis: This command needs to take place after the login mapping is applied. So we'd need to change order of tasks in tasks/main.yml.

examples/confined_users-playbook.yml
vmojzis: It's an example, which is why I'm trying to show both ways. Anyone using the playbook can just set `privileged_users: []`.

vmojzis:
> Why not include this into the main README.md?

The intention was to not "force" people who only want a simple way for confining users to read the whole "readme" (`confined_users-playbook.yml` uses a subset of SELinux system role's capabilities). Also, `privileged_users` variable is specific to  `confined_users-playbook.yml`.

examples/confined_users-playbook.yml
vmojzis: Ok, I guess `root` can stay "unconfined" since it's not accessible from outside. I'll replace it with a better example.

spetrosi:
> > Why not include this into the main README.md?
> 
> The intention was to not "force" people who only want a simple way for confining users to read the whole "readme" (`confined_users-playbook.yml` uses a subset of SELinux system role's capabilities). Also, `privileged_users` variable is specific to `confined_users-playbook.yml`.

The wording of `You can greatly improve security of your systems` sounds to me like a best practice for users to do. We can also restructure the README.md to split it into "scenarios" one of which would be `Configuring confined users`, like I did in mssql role https://linux-system-roles.github.io/mssql/. @bachradsusi WDYT?
We can do this in a future PR though, after the section is merged here as a separate readme.

bachradsusi:
Looking at https://linux-system-roles.github.io/selinux/ and https://linux-system-roles.github.io/mssql/ I guess it makes sense to merge "README-confined_users.md" into "README.md" as a new scenario.

Generally I like the idea to have confined users documented, but I think that the current approach unnecessarily mixes everything into one playbook. Also using `privileged_users` in the README is confusing for me as it's not related to the role, it's only implemented in the example playbook.

In the example playbook I'd also avoid all tasks which could be done by the role:

- instead of `shell: restorecon ...` use `selinux_restore_dirs` variable
- do not create users with assigned SELinux users, if you need to create a user, do it without SELinux part and demonstrate change of SELinux login mappings using `selinux_logins`

btw `user` module would not change SELinux user mappings if the user already exists

```
# ansible -i localhost, -c local -m user -a 'name=user_u seuser=user_u' all
localhost | CHANGED => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": true,
    "comment": "",
    "create_home": true,
    "group": 1002,
    "home": "/home/user_u",
    "name": "user_u",
    "shell": "/bin/bash",
    "state": "present",
    "system": false,
    "uid": 1002
}
# semanage login -l | grep user_u
user_u               user_u               s0                   *
# ansible -i localhost, -c local -m user -a 'name=user_u seuser=unconfined_u' all
localhost | SUCCESS => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "append": false,
    "changed": false,
    "comment": "",
    "group": 1002,
    "home": "/home/user_u",
    "move_home": false,
    "name": "user_u",
    "shell": "/bin/bash",
    "state": "present",
    "uid": 1002
}
# semanage login -l | grep user_u
user_u               user_u               s0                   *
```

vmojzis:
> Looking at https://linux-system-roles.github.io/selinux/ and https://linux-system-roles.github.io/mssql/ I guess it makes sense to merge "README-confined_users.md" into "README.md" as a new scenario.
> 
> Generally I like the idea to have confined users documented, but I think that the current approach unnecessarily mixes everything into one playbook. Also using `privileged_users` in the README is confusing for me as it's not related to the role, it's only implemented in the example playbook.

It felt better to me to document the variable in a separate README than to leave it out. Yes, it is only used in the example playbook, but is important for anyone utilizing the example.

> 
> In the example playbook I'd also avoid all tasks which could be done by the role:

> 
> * instead of `shell: restorecon ...` use `selinux_restore_dirs` variable

`restorecon` needs to be run after the login mapping is applied. So we'd need to change order of tasks in tasks/main.yml for this to work properly.

> * do not create users with assigned SELinux users, if you need to create a user, do it without SELinux part and demonstrate change of SELinux login mappings using `selinux_logins`
> 

Why? This is an example so I feel it is important to show multiple approaches.

> btw `user` module would not change SELinux user mappings if the user already existed
> 
> ```
> # ansible -i localhost, -c local -m user -a 'name=user_u seuser=user_u' all
> localhost | CHANGED => {
>     "ansible_facts": {
>         "discovered_interpreter_python": "/usr/bin/python3"
>     },
>     "changed": true,
>     "comment": "",
>     "create_home": true,
>     "group": 1002,
>     "home": "/home/user_u",
>     "name": "user_u",
>     "shell": "/bin/bash",
>     "state": "present",
>     "system": false,
>     "uid": 1002
> }
> # semanage login -l | grep user_u
> user_u               user_u               s0                   *
> # ansible -i localhost, -c local -m user -a 'name=user_u seuser=unconfined_u' all
> localhost | SUCCESS => {
>     "ansible_facts": {
>         "discovered_interpreter_python": "/usr/bin/python3"
>     },
>     "append": false,
>     "changed": false,
>     "comment": "",
>     "group": 1002,
>     "home": "/home/user_u",
>     "move_home": false,
>     "name": "user_u",
>     "shell": "/bin/bash",
>     "state": "present",
>     "uid": 1002
> }
> # semanage login -l | grep user_u
> user_u               user_u               s0                   *
> ```

bachradsusi:
> It felt better to me to document the variable in a separate README than to leave it out. Yes, it is only used in the example playbook, but is important for anyone utilizing the example.

It depends on what's the goal. If the goal a generic SELinux example using Ansible SELinux modules and creating users then it's probably ok. But I'd expect that README and examples in linux-system-roles would be related to roles. I understand you need to create sudoers so that administrator users could switch to sysadm_t using sudo, but I don't think it's necessary to complicate example playbook with a new list variable and use loops. It would be enough to demonstrate the change on one hardcoded user which is documented as pre-created.
 
> > In the example playbook I'd also avoid all tasks which could be done by the role:
> 
> > * instead of `shell: restorecon ...` use `selinux_restore_dirs` variable
> 
> `restorecon` needs to be run after the login mapping is applied. So we'd need to change order of tasks in tasks/main.yml for this to work properly.

You are right. It's a bug in the role and the order needs to be changed. restorecon should be the last step.


> > * do not create users with assigned SELinux users, if you need to create a user, do it without SELinux part and demonstrate change of SELinux login mappings using `selinux_logins`
> 
> Why? This is an example so I feel it is important to show multiple approaches.

It also depends on the goal. Using `user` module with `seuser` option is not related to the role.


Anyway, even though I don't agree with some parts of this PR, I know that @vmojzis is experienced SELinux engineer and so if other maintainers approve this PR I'm not going to block it.

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+      - "Lizzy"
+      - "Tarja"
+
+    # The following section specifies mapping of EXISTING system users
+    # to SELinux users
+    selinux_logins:
+      # Map all new users to SElinux user "user_u" (instead of "unconfined_u")
+      - {login: '__default__', seuser: 'user_u', serange: 's0',
+         state: 'present'}
+      # Alternatively map all new users to SElinux user "staff_u" (instead of
```
richm, line 49:
```suggestion
        # Alternatively map all new users to SElinux user "staff_u" (instead of
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+      # "unconfined_u")
```
richm, line 50:
```suggestion
        # "unconfined_u")
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+      #- {login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023',
+      #   state: 'present'}
+      # Map admin users to SElinux user "sysadm_u"
+      # Map select non-admin privileged users to SELinux user "staff_u"
+      # - {login: '<existing_user>', seuser: 'staff_u', serange: 's0',
+      #    state: 'present'}
+      #- {login: 'Lizzy', seuser: 'staff_u', serange: 's0', state: 'present'}
```
richm, line 57:
```suggestion
      # - {login: 'Lizzy', seuser: 'staff_u', serange: 's0', state: 'present'}
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+
+  # Prepare the prerequisites required for this playbook
+  tasks:
+    # Enable kiosk mode -- add extremely limited users "guest" and "xguest"
+    #- name: Install xguest package
```
richm, line 62:
```suggestion
    # - name: Install xguest package
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+    #  yum:
+    #    name: xguest
+    #    state: present
+
+    # Create login mappings for "existing_privileged_users"
+    - set_fact:
```
richm, line 68:
```suggestion
    - name: Build list of selinux logins
      set_fact:
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+        selinux_logins: "{{ selinux_logins + login_helper | list }}"
+      with_items: "{{ existing_privileged_users }}"
+      vars:
+        login_helper:
+          - {login: "{{ item }}", seuser: 'staff_u', serange: 's0',
+             state: 'present'}
+
+    # Add new system users that will be assigned to staff_u SELinux user
+    - name: Add non-root privileged users
+      user:
```
richm, line 78: I guess this assumes you aren't using IdM or some other system that you must create users in

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+            msg: "role failed"
+          when: not selinux_reboot_required
+
+        - name: Restart managed host
+          reboot:
+
+        - name: Wait for managed host to come back
+          wait_for_connection:
+            delay: 10
+            timeout: 300
```
richm, line 109: why so long?  Is this because selinux has to do some sort of filesystem-wide restorecon?

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+            name: linux-system-roles.selinux
+
+    # Users that got assigned a new SELinux user need to have their homedirs
+    # relabeled.
+    # Gather homedirs of each user with a new login mapping
+    - name: Gather homedirs of users with new login mapping
+      shell: >
+        getent passwd "{{ item.login }}" | awk -F: '{ print $6 }'
+      register: __login_homedirs
+      with_items: "{{ selinux_logins }}"
```
richm, line 122:
```suggestion
      with_items: "{{ selinux_logins }}"
      changed_when: false
```

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+
+        - name: Reapply the role
+          include_role:
+            name: linux-system-roles.selinux
+
+    # Users that got assigned a new SELinux user need to have their homedirs
+    # relabeled.
+    # Gather homedirs of each user with a new login mapping
+    - name: Gather homedirs of users with new login mapping
+      shell: >
```
richm, line 119:
```suggestion
      shell: >-
        set -euo pipefail;
```

richm:
> You are right. It's a bug in the role and the order needs to be changed. restorecon should be the last step.

Does this mean the role can't manage confined users until this is fixed?

bachradsusi:
> Does this mean the role can't manage confined users until this is fixed?

The user part of the file context - unconfined_u, staff_u, ... - is not used for access decisions in Fedora policies so `restorecon -F` is not actually necessary. 

```
$ sestatus  | grep policy\ name
Loaded policy name:             targeted
$ id -Z
user_u:user_r:user_t:s0
$ ls -Z date
unconfined_u:object_r:user_home_t:s0 date
$ cat date
Tue Aug 29 09:59:34 AM EDT 2023
```

It could trigger security scan warnings though.

vmojzis:
Sorry for the delay.
I dropped the use of `user` module, fixed relabeling (it is now the last task in the role and the example uses `selinux_restore_dirs`) and incorporated all the stylistic tweaks.

examples/confined_users-playbook.yml
vmojzis: This is just copied from the first example.

examples/confined_users-playbook.yml
vmojzis: Fixed, thanks (aside from the "-o pipefail").

examples/confined_users-playbook.yml
vmojzis: I skipped this one since it would indent the comment from the surrounding lines.

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+
+    # The following section specifies mapping of EXISTING system users
+    # to SELinux users
+    selinux_logins:
+      # Map all new users to SElinux user "user_u" (instead of "unconfined_u")
+      - {login: '__default__', seuser: 'user_u', serange: 's0',
+         state: 'present'}
+      # Alternatively map all new users to SElinux user "staff_u" (instead of
+      # "unconfined_u")
+      #- {login: '__default__', seuser: 'staff_u', serange: 's0-s0:c0.c1023',
```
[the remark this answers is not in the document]
vmojzis: Fixed, thanks.

examples/confined_users-playbook.yml
vmojzis: Fixed, thanks.

examples/confined_users-playbook.yml
vmojzis: Fixed, thanks.

examples/confined_users-playbook.yml
vmojzis: Yes, that seemed like too complex a task to have in an example.

examples/confined_users-playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+    # Gather homedirs of each user with a new login mapping
+    - name: Gather homedirs of users with new login mapping
+      shell: >
+        getent passwd "{{ item.login }}" | awk -F: '{ print $6 }'
+      register: __login_homedirs
+      with_items: "{{ selinux_logins }}"
+    # Relabel the gathered homedirs (-F is necessary to change the user portion
+    # of file labels)
+    - name: Relabel homedirs of users with new login mapping
+      shell: "restorecon -RF {{ item.stdout }}"
```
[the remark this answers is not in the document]
vmojzis: Removed (I used `selinux_restore_dirs` instead ).

README-confined_users.md
```diff
@@ -0,0 +1,59 @@
+# Confined users
```
richm, line 1:
```suggestion
# Confined users

```

README-confined_users.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Expected functionality
+
+* Set default user mapping
+* Confine existing users accounts
+* Set booleans to customize access
+
+## Variables
+
+### selinux_booleans
```
richm, line 13:
```suggestion
### selinux_booleans

```

examples/confined_users-playbook.yml
richm: ok - what do you suggest we do to make yamllint happy?

README-confined_users.md
vmojzis: Fixed, thank you.

README-confined_users.md
vmojzis: Fixed, thank you.

richm:
@bachradsusi is this PR ok to merge?

vmojzis:
Fixed some ansible-lint issues (but I cannot fix the "pipefail" and do not agree with the comment indentation on line 44).

bachradsusi:
> @bachradsusi is this PR ok to merge?

yes

richm (approved):

PR #189, opened 2023-09-12
Fix #188

### LINKED ISSUE #188: defined selinux_ports cause error when selinux_state=disabled
Setting `selinux_ports` will cause the module to fail if  `selinux_state=disabled`. We require to set `selinux_ports` globally.

Workaround: Setting `selinux_ports=[]` for all hosts that `selinux_state=disabled` and hence bloating the inventory.

Is there a way to avoid this without PR?

goshansp:
Set an SELinux label on a port only when selinux_state is not disabled

Enhancement: Fix #188

Reason: Avoid bloating inventory for hosts that have `selinux_state=disabled`

Result: ?

Issue Tracker Tickets (Jira or BZ if any): https://github.com/linux-system-roles/selinux/issues/188

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -103,6 +103,7 @@
     state: "{{ item.state | default('present') }}"
     local: "{{ item.local | default(False) }}"
   with_items: "{{ selinux_ports }}"
+  when: selinux_state != "disabled"
 
 - name: Set linux user to SELinux user mapping
   selogin:


richm:
[citest]

richm:
Thank you for the PR.

Please also update the README.md - add something like "If `selinux_state` is `disabled`, then ports will not be managed.".

Do we need to do something similar for selinux types other than ports?  e.g. if selinux_state disabled, should we skip managing selinux_booleans, selinux_fcontexts, selinux_logins, and selinux_modules?

bachradsusi:
Thanks.

But I don't think this is the right approach when you compare results with `semanage port` in SELinux disabled systems.

- when "selinux-policy-targeted" is installed `semanage port` works as expected:

```
[root@fedora ~]# sestatus 
SELinux status:                 disabled
[root@fedora ~]# rpm -q selinux-policy-targeted
selinux-policy-targeted-38.27-1.fc40.noarch
[root@fedora ~]# semanage port -a -t ssh_port_t -p tcp 2230
[root@fedora ~]# semanage port -l | grep 2230
ssh_port_t                     tcp      2230, 2224, 2223, 2222, 22
```

- without "selinux-policy-targeted" `semanage port` generates error:

```
[root@fedora ~]# rpm -q selinux-policy-targeted
package selinux-policy-targeted is not installed
[root@fedora ~]# sestatus 
SELinux status:                 disabled
[root@fedora ~]# semanage port -a -t ssh_port_t -p tcp 2231
libsemanage.semanage_read_policydb: Could not open kernel policy /var/lib/selinux/targeted/active/policy.kern for reading. (No such file or directory).
FileNotFoundError: No such file or directory
```

So it's not generally related to status of SELinux but to the state of filesystem - existence of "/etc/selinux/config"; or "/etc/selinux/targeted" and "/var/linux/selinux/targeted".

It's expected that a system can be configured before SELinux is enabled.

goshansp:
Thank you for looking into it and providing valuable insight. How can we solve #188 ? Should we change the title of this PR to one of the following?

- Install selinux-policy-targeted if sestatus=disabled
- Print warning for missing selinux-policy-targeted when sestatus=disabled

bachradsusi:
>  *  Install selinux-policy-targeted if sestatus=disabled

there's no reason to install "selinux-policy-targeted" on SELinux disabled system

> *  Print warning for missing selinux-policy-targeted when sestatus=disabled

I'm thinking about dropping `get_runtime_status()` - https://github.com/linux-system-roles/selinux/blob/main/library/local_seport.py#L312 and better handling of exception coming from `seobject.portRecords`, e.g. provide warning and report no change if SELinux is disabled and there's no "targeted" policy in the system.

But I need some time to experiment with this.

Anyway, I'd close this PR and move the discussion to the https://github.com/linux-system-roles/selinux/issues/188 issue.

PR #194, opened 2023-09-20
Use `ignore_selinux_state` module option

bachradsusi:
It is possible to configure SELinux on nodes with SELinux disabled before they're switch to permissive/enforcing. The only requirement is that targeted SELinux policy (or a policy configured in /etc/selinux/config) needs to be installed.

Fixes: https://github.com/linux-system-roles/selinux/issues/188

Note that boolean settings doesn't work work even with `ignore_selinux_state=true` - it's a bug in seboolean module which uses `selinux.security_get_boolean_names()` instead of `seobject` API from semanage.

Enhancement:

Reason:

Result:

Issue Tracker Tickets (Jira or BZ if any):


spetrosi:
[citest]

spetrosi:
lgtm, let's wait for tests. Please update PR description by filling in Enhancement:, Reason:, Result: template.

bachradsusi:
Please hold this PR for now. I'll probably make another change booleans settings, Something like the following change could improve the role for SELinux disabled mode. I'm working also on seboolean module change for this.

```
diff --git a/tasks/main.yml b/tasks/main.yml
index 5cd0ee23efda..9c35a5ecb8a3 100644
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -68,7 +68,8 @@
   seboolean:
     name: "{{ item.name }}"
     state: "{{ item.state }}"
-    persistent: "{{ item.persistent | default('no') }}"
+    persistent: "{{ item.persistent \
+                  | default( ansible_selinux.status == 'disabled' ) }}"
     ignore_selinux_state: "{{ ansible_selinux.status == 'disabled' }}"
   with_items: "{{ selinux_booleans }}"
 
```

bachradsusi:
seboolean PR - https://github.com/ansible-collections/ansible.posix/pull/496

bachradsusi:
The PR is updated and if everything pass I think it's ready. It'll be possible to set SELinux booleans when https://github.com/ansible-collections/ansible.posix/pull/496 is accepted and shipped.

tasks/main.yml
```diff
@@ -68,7 +68,9 @@
   seboolean:
     name: "{{ item.name }}"
     state: "{{ item.state }}"
-    persistent: "{{ item.persistent | default('no') }}"
+    persistent: "{{ item.persistent \
```
richm, line 71:
```suggestion
    persistent: "{{ item.persistent |
```

tasks/main.yml
```diff
@@ -68,7 +68,9 @@
...
+                  | default( ansible_selinux.status != 'disabled' ) }}"
```
richm, line 72:
```suggestion
                  default( ansible_selinux.status != 'disabled' ) }}"
```

richm:
How can we test this?  Do we already have a test that attempts to disable selinux then update settings?
Do we need to update the README.md for this?
Can we test this before https://github.com/ansible-collections/ansible.posix/pull/496 is merged?
Do we need to change the role after https://github.com/ansible-collections/ansible.posix/pull/496 is merged?

bachradsusi:
> How can we test this?

```
  vars:
    selinux_booleans:
       - {name: 'ssh_sysadm_login', state: 'off', persistent: 'yes'}
    selinux_fcontexts:
      - {target: '/tmp/test_dir(/.*)?', setype: 'user_home_dir_t', ftype: 'd'}
    selinux_ports:
      - {ports: '22100', proto: 'tcp', setype: 'ssh_port_t', state: 'present'}
    selinux_logins:
      - {login: 'sar-user', seuser: 'staff_u', serange: 's0-s0:c0.c1023',
         state: 'present'}
```

on SELinux disabled system

You could simulate SELinux disabled system using `umount -l /sys/fs/selinux`

>Do we need to update the README.md for this?

Maybe add a note that it's possible to configure SELinux in disabled mode when selinux-policy-targeted is installed.

> Can we test this before [ansible-collections/ansible.posix#496](https://github.com/ansible-collections/ansible.posix/pull/496) is merged?

All but booleans should work without [ansible-collections/ansible.posix#496](https://github.com/ansible-collections/ansible.posix/pull/496)

> Do we need to change the role after [ansible-collections/ansible.posix#496](https://github.com/ansible-collections/ansible.posix/pull/496) is merged?

No

tasks/main.yml
```diff
@@ -68,7 +68,9 @@
   seboolean:
     name: "{{ item.name }}"
     state: "{{ item.state }}"
-    persistent: "{{ item.persistent | default('no') }}"
+    persistent: "{{ item.persistent |
+                  default( ansible_selinux.status != 'disabled' ) }}"
```
richm, line 72:
```suggestion
                  default(ansible_selinux.status != 'disabled') }}"
```

bachradsusi:
> > Do we need to update the README.md for this?
> 
> Maybe add a note that it's possible to configure SELinux in disabled mode when selinux-policy-targeted is installed.

Note added

tasks/main.yml
```diff
@@ -68,7 +68,9 @@
   seboolean:
     name: "{{ item.name }}"
     state: "{{ item.state }}"
-    persistent: "{{ item.persistent | default('no') }}"
+    persistent: "{{ item.persistent |
+                  default( ansible_selinux.status == 'disabled' ) }}"
```
richm, line 72:
```suggestion
                  default(ansible_selinux.status == 'disabled') }}"
```

tasks/main.yml
```diff
@@ -68,7 +68,9 @@
   seboolean:
     name: "{{ item.name }}"
     state: "{{ item.state }}"
-    persistent: "{{ item.persistent | default('no') }}"
+    persistent: "{{ item.persistent |
+                  default(ansible_selinux.status == 'disabled') }}"
```
richm, line 72: This changes the data type, and the data type will be different - the result of `ansible_selinux.status == 'disabled'` will be `true` or `false` of type `bool` - can the `seboolean` module take either a string `"yes"`/`"no"` value, or a boolean `true`/`false` value?

bachradsusi: Does it matter? `persistent` option is `bool` and according to https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#boolean-variables `bool` could be both `true` or `'yes'`

richm: ok - looks like https://github.com/ansible-collections/ansible.posix/blob/main/plugins/modules/seboolean.py#L278 does convert bool-ish strings to `bool`

richm (approved):

PR #195, opened 2023-09-26
fix: make role work again on Suse - not officially supported

richm:
Note that Suse is unofficially supported.  The recent change for
the python3 library names made the role break on Suse.


richm:
[citest]

PR #206, opened 2023-10-06
feat: support for ostree systems

richm:
Feature: Allow running and testing the role with ostree managed nodes.

Reason: We have users who want to use the role to manage ostree
systems.

Result: Users can use the role to manage ostree managed nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,18 +17,16 @@ Essentially provide mechanisms to manage local customizations:
 
 ## Requirements
 
+See below
+
 ### Collection requirements
 
-The role requires some SELinux modules.  If you are using `ansible-core`, you
-must get these from the `ansible.posix` and `community.general` collections.
-Use the file `meta/collection-requirements.yml` to install these:
+The role requires external collections.  Use the following command to install
+them:
 
 ```bash
 ansible-galaxy collection install -vv -r meta/collection-requirements.yml
 ```
-
-If you are using Ansible Engine 2.9, or are using an Ansible bundle which
-includes these collections/modules, you should have to do nothing.
 
 ## Modules provided by this repository
 
@@ -231,3 +229,7 @@ on Red Hat Enterprise Linux 6
 
 The general usage is demonstrated in
 [selinux-playbook.yml](examples/selinux-playbook.yml) playbook.
+
+## rpm-ostree
+
+See README-ostree.md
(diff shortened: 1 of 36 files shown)


richm:
[citest]

richm:
[citest]

richm:
[citest]

README.md
```diff
@@ -231,3 +229,71 @@ on Red Hat Enterprise Linux 6
 
 The general usage is demonstrated in
 [selinux-playbook.yml](examples/selinux-playbook.yml) playbook.
+
+## rpm-ostree
```
spetrosi, line 233: We can put this info into a separate .md file the same way we do with [e.g.](https://github.com/linux-system-roles/.github/blob/main/playbooks/templates/README-ansible.md).  It'd would be nice to keep such a specific info not in the main README, WDYT?

richm: That would make it much easier to keep it up-to-date, and not "pollute" the file with information not needed by users who aren't interested in ostree.  That brings up another point - we'll need to document this somehow in the collection https://galaxy.ansible.com/ui/repo/published/fedora/linux_system_roles/docs/ - we'll either need to add this to the collection readme, or have a separate file there too e.g. README-ostree

@briansmith0 wdyt?

spetrosi: If you look at rhel-system-roles AH page, the main README.md is printed on the main page of the collection. Then, there is a **Documentation** tab where you can find all other md files, now they are roles' READMEs and CHANGELOGs. For visibility, it's better to put info about ostree is to put it to the main collection README.

richm: @spetrosi @briansmith0 please review the latest README.md and the new README-ostree.md

briansmith0: > @spetrosi @briansmith0 please review the latest README.md and the new README-ostree.md

Those files look good to me.  Thanks!

richm:
[citest]

README.md
richm: done.  And I will be adding this to https://github.com/linux-system-roles/.github along with some other changes for ostree

richm:
[citest]

richm:
[citest]

richm:
@spetrosi any objection to merging this?  I need to first publish the roles that are used by other roles before working on those other roles

Issue #216, opened 2023-12-07
[WARNING]: TASK: ... "The loop variable 'item' is already in use." is displayed when calling the role from another role


berndfinger:
When calling the `selinux` role from another role and looping over variables, as in https://github.com/sap-linuxlab/community.sap_install/blob/5b920752aa818fcd9fe96ce740bffaa314baf99e/roles/sap_general_preconfigure/tasks/RedHat/generic/configure-selinux.yml#L108 (the task is looping over the list `sap_general_preconfigure_sap_directories`, which is assigned to the selinux variable `selinux_restore_dirs`), messages like the following are displayed:
```
[WARNING]: TASK: fedora.linux_system_roles.selinux : Restore SELinux labels on filesystem tree: The loop variable 'item' is already in use.
You should set the `loop_var` value in the `loop_control` option for the task to something else to avoid variable collisions and unexpected behavior.
```
As indicated in the warning message, we can avoid this message by replacing `item` in the `selinux` role with something else, e.g. `line_item`, and also adding:
```
  loop_control:
    loop_var: line_item
```
to the affected tasks.

PR #217, opened 2023-12-07
fix: no longer use "item" as a loop variable

berndfinger:
By using a different loop variable, we can avoid WARNING messages related to loop variables when calling the `selinux` role from another role.

Fixes: https://github.com/linux-system-roles/selinux/issues/216 .

Enhancement: Avoid the warning message:
```
[WARNING]: TASK: fedora.linux_system_roles.selinux : Restore SELinux labels on filesystem tree: The loop variable 'item' is already in use.
You should set the `loop_var` value in the `loop_control` option for the task to something else to avoid variable collisions and unexpected behavior.
```

Reason: Although I am not aware of any errors when using the role `selinux` without this fix, the warning message (printed in bold and in a different than usual color - violet in my tests) adds lines to the output which can confuse users.

Result: The warning message `[WARNING]: TASK: ... "The loop variable 'item' is already in use." is no longer displayed when calling the role from another role.

Issue Tracker Tickets (Jira or BZ if any): N/A

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -66,44 +66,52 @@
 - name: Set SELinux booleans
   # noqa fqcn[action]
   seboolean:
-    name: "{{ item.name }}"
-    state: "{{ item.state }}"
-    persistent: "{{ item.persistent |
+    name: "{{ line_item.name }}"
+    state: "{{ line_item.state }}"
+    persistent: "{{ line_item.persistent |
                   default(ansible_selinux.status == 'disabled') }}"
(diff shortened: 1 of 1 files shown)


richm:
[citest]

richm:
These test failures are really strange - I don't think they are related to the PR - investigating

richm:
@berndfinger in tasks/selinux_load_module.yml `item` needs to be renamed to `__selinux_item`

richm:
[citest]

richm (approved):

PR #218, opened 2023-12-08
fix: Print an error message when module to be created doesn't exist

spetrosi:
Enhancement: Print an error message when module to be created doesn't exist

Reason: When specifying a non-existent file with `selinux_modules.path` the role fails with an informative message that the module file that you try to install doesn't exist.

### DIFF
--- a/tasks/selinux_load_module.yml
+++ b/tasks/selinux_load_module.yml
@@ -26,6 +26,13 @@
       delegate_to: localhost
       become: false
 
+    - name: Raise an error when module file doesn't exist
+      fail:
+        msg: >-
+          The module that you try to install does not exist in
+          {{ __selinux_item.path }}
+      when: not module_file.stat.exists
+
     - name: Install module
       when: not module_file.stat.checksum in checksum
       block:


tasks/selinux_load_module.yml
```diff
@@ -27,7 +27,9 @@
       become: false
 
     - name: Install module
-      when: not module_file.stat.checksum in checksum
+      when: >-
+        (not module_file.stat.exists) or
+        (module_file.stat.exists and not module_file.stat.checksum in checksum)
```
richm, line 32: while this seems correct, unfortunately it will break https://github.com/linux-system-roles/selinux/blob/main/tests/tests_selinux_modules.yml#L149 because an error won't be raised by this code.  I think we should not fix this.  Or, if we do, we need to figure out if it is ok to always raise an error if `not module_file.stat.exists`

richm:
[citest]

tasks/selinux_load_module.yml
```diff
@@ -26,6 +26,11 @@
       delegate_to: localhost
       become: false
 
+    - name: Raise an error when module file doesn't exist
+      fail:
+        msg: The module that you try to install doesn't exist in {{ item.path }}
```
richm, line 31:
```suggestion
        msg: The module that you try to install doesn't exist in {{ __selinux_item.path }}
```

richm:
[citest]

richm:
the rhel6 and 7 failures are known and seen occasionally
the fedora 39 error looks odd, possibly transient - I don't see it locally

PR #219, opened 2023-12-08
docs: Update docs to use yaml style when defining vars

spetrosi:
Enhancement: Update docs to use yaml style when defining vars

Reason: The role was using an old jinja style definitions

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,11 @@ module.
 
 ```yaml
 selinux_booleans:
-  - {name: 'samba_enable_home_dirs', state: true}
-  - {name: 'ssh_sysadm_login', state: true, persistent: true}
+  - name: samba_enable_home_dirs
+    state: true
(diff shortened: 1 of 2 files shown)


richm:
some background - constructs like `varname: {key: value, ...}` are JSON "embedded" in YAML.  This is generally discouraged in Ansible - https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style "Even if JSON is valid YAML and Ansible understands it, do only use JSON syntax if it makes sense (e.g. a variable file automatically generated) or adds to the readability. In doubt, nobody expects JSON so stick to YAML."

However, in some cases, it can make the code more readable e.g.

```yaml
list_var: [a, b, c, d]
```
is imo more readable than
```yaml
list_var:
  - a
  - b
  - c
  - d
```
for short lists with short list members

PR #267, opened 2025-04-08
ci: several changes related to new qemu test, ansible-lint, python versions, ubuntu versions

richm:
There is a new QEMU based test which uses the qemu/kvm capability of
github action runners.  This is the basis for new bootc/image mode tests
which we will be rolling out in the near future.

ansible-lint requires that the collection path is set so that the requirements
it installs are installed in the correct place.

There has been some general github action deprecation of python versions and
ubuntu versions that we have had to fix.

Remove `CONTRIBUTOR` from the list of users who can trigger citest.

For more information, see

* https://github.com/linux-system-roles/.github/pull/98
* https://github.com/linux-system-roles/.github/pull/94
* https://github.com/linux-system-roles/.github/pull/93
* https://github.com/linux-system-roles/.github/pull/92
* https://github.com/linux-system-roles/.github/pull/91


martinpitt:
[citest]

martinpitt (approved):

PR #285, opened 2025-06-21
test: use daemons_use_tcp_wrapper for seboolean tests

richm:
daemons_use_tcp_wrapper is supported on all platforms including
automotive.


richm:
[citest]

PR #294, opened 2025-08-12
test: restore selinux config file in proper context

xjezda00:
Use _default SELinux level, role, type and user when copy into /etc/selinux/config for ensuring that file has proper SELinux context.

Enhancement:

Reason:
Solves AVCs after tests_selinux_disabled.yml is executed

Result:
No AVCs

Issue Tracker Tickets (Jira or BZ if any):


milosmalik:
Looks good to me :+1:

richm (approved):

PR #345, opened 2026-04-27
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.

## Summary by Sourcery

Add fingerprint logging for the SELinux system role and verify it via system journal tests.

New Features:
- Introduce an sr_fingerprint Ansible module to write structured fingerprint messages to syslog with timestamps.
- Record role begin and success fingerprints for the SELinux system role including Ansible version and platform information.

CI:
- Add Ansible sanity ignore configuration files for multiple Ansible versions to accommodate the new custom module.

Tests:
- Extend the default role test playbook to verify that the expected begin and success fingerprint entries appear in the system journal when syslog is available.

### DIFF
--- a/.sanity-ansible-ignore-2.14.txt
+++ b/.sanity-ansible-ignore-2.14.txt
@@ -1,0 +1,1 @@
+plugins/modules/sr_fingerprint.py validate-modules:missing-gplv3-license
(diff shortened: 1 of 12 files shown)


richm:
[citest]

Issue #352, opened 2026-05-20
restorecon invalid argument on first run when a module defines a type referenced by selinux_restore_dirs


ozon2:
When the role installs a custom SELinux policy module and runs `restorecon` over a path covered by file contexts defined in that module's `.fc` file in the same play, the run fails on first apply with:
 
```
restorecon: Could not set context for <file>: Invalid argument
```

Subsequent runs succeed.

As I am a beginner with selinux, please tell me if this is not the standard way of doing this.

## Environment

- fedora.linux_system_roles.selinux v1.125.1
- RHEL 9
- ansible-core 2.16

## Example

A minimal policy module that declares a new file type and labels a directory:

```
# policy.te
policy_module(policy, 1.0.0)

type policy_var_lib_t;
files_type(policy_var_lib_t)
```

```
# policy.fc
/opt/policy/data(/.*)?    gen_context(system_u:object_r:policy_var_lib_t,s0)
```

Build with `make -f /usr/share/selinux/devel/Makefile policy.pp` on the target, copy `policy.pp` next to the playbook.

```yaml
- hosts: target
  become: true
  tasks:
    - name: Create the labeled directory and a file inside it
      file: { path: /opt/policy/data, state: directory }
    - copy: { content: "hello", dest: /opt/policy/data/test.txt }

    - name: Apply SELinux policy
      include_role:
        name: fedora.linux_system_roles.selinux
      vars:
        selinux_state: enforcing
        selinux_modules:
          - { path: {{ playbook_dir }}/policy.pp, state: enabled }
        selinux_restore_dirs:
          - /opt/policy/data
```

### Observed

```
TASK [fedora.linux_system_roles.selinux : Install the module] ******************
changed: [target]

TASK [fedora.linux_system_roles.selinux : Clean temporary directory ...] *******
changed: [target]

TASK [fedora.linux_system_roles.selinux : Refresh SELinux modules facts] *******
ok: [target]

TASK [fedora.linux_system_roles.selinux : Set SELinux modules facts] ***********
ok: [target]

TASK [fedora.linux_system_roles.selinux : Restore SELinux labels on filesystem tree] ***
failed: [target] (item=/opt/policy/data) => {
    "cmd": ["/sbin/restorecon", "-R", "-F", "-v", "-T", "0", "/opt/policy/data"],
    "msg": "non-zero return code",
    "rc": 255,
    "stderr_lines": [
[...]

PR #354, opened 2026-06-04
fix: ensure policy modules are loaded before they can be referenced

richm:
Cause: Policy modules were not being loaded with `semodule -R` after being changed.

Consequence: If other policy referred to the policy in the module, then the
subsequent `restorecon` would fail when it tried to reference a policy that was
not loaded.

Fix: Any policy change will trigger a `semodule -R` before doing the `restorecon`.

Result: User can apply custom policy modules and other policy that references the
custom policy without errors.

## Summary by CodeRabbit

* **Tests**
  * Added a new SELinux test policy module and associated file contexts.
  * Expanded test scenarios and setup/cleanup for Fedora/EL7 and EL6, including instructions to rebuild tests on older OS.

* **Refactoring**
  * Improved SELinux policy reload handling by explicitly tracking when module operations change policy and only triggering a reload when needed and SELinux is active.


richm:
fixes https://github.com/linux-system-roles/selinux/issues/352

richm:
[citest]

spetrosi (approved):

bachradsusi:
In general works for me from SELinux point of view.

I'm not sure I understand why `__selinux_modules_needs_reload` is set as fact and what all these `__selinux_policy_installed is defined and __selinux_policy_installed is changed` are good for.

bachradsusi (approved):

richm:
> In general works for me from SELinux point of view.
> 
> I'm not sure I understand why `__selinux_modules_needs_reload` is set as fact and what all these `__selinux_policy_installed is defined and __selinux_policy_installed is changed` are good for.

The current implementation with the handler notifies the handler every time something changes:

* a module is installed https://github.com/linux-system-roles/selinux/blob/main/tasks/selinux_load_module.yml#L57
* a module is removed https://github.com/linux-system-roles/selinux/blob/main/tasks/selinux_load_module.yml#L72
* a module is enabled https://github.com/linux-system-roles/selinux/blob/main/tasks/selinux_load_module.yml#L79
* a module is disabled https://github.com/linux-system-roles/selinux/blob/main/tasks/selinux_load_module.yml#L86

Since selinux_load_module.yml is called in a loop, we don't want to run `semodule -R` for every loop iteration. The handler takes care of that, only running `semodule -R` once no matter how many module changes there are.  But we cannot wait for the handler to run, we need to call `semodule -R` immediately after making any module related changes (or, at least, before calling `restorecon`), and calling `meta: flush_handlers` would trigger *ALL* pending handlers in the play, not just in the selinux role, so we have to manually recreate what the handler is doing:

* before the loop that calls selinux_load_module.yml for each module, set `__selinux_modules_needs_reload: false`
* for each task that previously called `notify:` for the handler, instead use `register:` to record the result of the task
* at the end of selinux_load_module.yml, ensure `__selinux_modules_needs_reload` is set to `true` if there were any changes
* after the loop that calls selinux_load_module.yml for each module, if `__selinux_modules_needs_reload` is `true`, then call `semodule -R`

PR #374, opened 2026-08-20
fix: Vendor ansible-2.9-compatible SELinux modules

spetrosi:
Enhancement:
Vendor ansible-2.9-compatible copies of the external collection modules into the role's library/ directory and duplicate each task that uses them: one variant uses the vendored module on RHEL 7, the other uses the latest upstream module everywhere else.

- selinux_ansible_29, seboolean_ansible_29 (from ansible.posix 2.1.0)
- sefcontext_ansible_29, selogin_ansible_29 (from community.general 11.4.9)
  The vendored modules replace the collection _respawn import with an inline no-op stub so they carry no module_utils dependency.

seport is left as the single existing local_seport module (used everywhere): it is already customized for dccp/sctp and is py2.7-compatible, so no ansible_29-specific variant is needed.

Remove the collection-requirements.yml version constraints (both the upper caps and the lower bounds) so modern systems install the latest ansible.posix and community.general.

Add "fromm" (author surname) to .codespell_ignores.

Reason:
The role must keep supporting RHEL 7 that runs ansible-2.9, but the external collection modules dropped support for the older Ansible shipped on RHEL 7 after the pinned versions.
Pinning the collections held every system back to those old releases.
Vendoring older copies lets RHEL 7 keep working while all other systems use the latest community modules.

Result:
ansible 2.9 hosts (e.g. RHEL 7) run the vendored ansible-2.9 modules; all other systems run the latest collection modules.
The two SELinux-state tasks use distinct _ansible_29 registers coalesced back into the original variable names to avoid a
skipped-task register clobber, so downstream reboot-required logic is unchanged.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

## Summary by CodeRabbit

- **New Features**
  - Added SELinux support for Ansible 2.9, including policy state, booleans, file contexts, and login mappings.
  - Added detection for Red Hat-family distributions running versions 7 through 10.
  - SELinux operations now support version-appropriate behavior, check mode, persistent settings, and reboot reporting.

- **Bug Fixes**
  - Improved compatibility with newer `ansible.posix` and `community.general` releases by removing restrictive version limits.

### DIFF
--- a/tests/tests_all_transitions.yml
+++ b/tests/tests_all_transitions.yml
@@ -26,6 +26,9 @@
     - name: Include test variables
       import_tasks: set_selinux_variables.yml
 
+    - name: Include RH distros variables
+      include_vars: vars/rh_distros_vars.yml
+
     - name: Save config
       include_tasks: selinux_config_save.yml
 
@@ -44,13 +47,30 @@
           ansible_facts['distribution_major_version'] is version('8', '>=') and
           ansible_version.full is version('2.10', '>=') }}"
 
-    # Bare name resolves to the ansible.posix builtin on EL7 (ansible 2.9) and
-    # redirects to the latest ansible.posix.selinux on newer systems.
+    # EL7 (ansible 2.9) uses the vendored selinux_el7 module; other systems use
+    # the bare selinux name, which redirects to the latest ansible.posix.selinux.
+    - name: Disable selinux and set kernel - EL7
+      selinux_el7:
+        state: disabled
+        update_kernel_param: "{{ true if __test_update_kernel_param else omit }}"
+      register: __disable_result_el7
+      when: __selinux_is_rh_distro_7
+
+    # Bare name (not FQCN): an unresolvable FQCN aborts play parsing on EL7 even
+    # when the task is skipped. The bare selinux name resolves to the ansible 2.9
+    # builtin (skipped here) and redirects to ansible.posix.selinux elsewhere.
     - name: Disable selinux and set kernel
       selinux:  # noqa fqcn
         state: disabled
         update_kernel_param: "{{ true if __test_update_kernel_param else omit }}"
-      register: __disable_result
+      register: __disable_result_non_el7
+      when: not __selinux_is_rh_distro_7
+
+    - name: Set disable result
+      set_fact:
+        __disable_result: "{{ __disable_result_el7
+          if __selinux_is_rh_distro_7
+          else __disable_result_non_el7 }}"
 
     - name: Add selinux=0 to kernel args
       when:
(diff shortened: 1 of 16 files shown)


spetrosi:
[citest]

spetrosi:
[citest]

spetrosi:
ansible 2.9 test will fail until https://github.com/linux-system-roles/tft-tests/pull/103 is merged. I'll wait until I have all PRs for this purpose ready for roles, then will merge test PR https://github.com/linux-system-roles/tft-tests/pull/103, then test the role PRs and merge them.

richm:
I suggest adding a variable to `set_facts_packages.yml` - `__selinux_use_el7_modules` - for other roles, it would be set in `tasks/set_vars.yml`

spetrosi:
[citest]

spetrosi:
[citest]

richm:
```
TASK [Run the role normally] ***************************************************
task path: /tmp/collections-QXS/ansible_collections/fedora/linux_system_roles/tests/selinux/tasks/run_role_with_clear_facts.yml:33
Friday 21 August 2026  14:53:46 -0400 (0:00:00.031)       0:00:08.369 ********* 
ERROR! couldn't resolve module/action 'ansible.posix.selinux'. This often indicates a misspelling, missing collection, or incorrect module path.

The error appears to be in '/tmp/collections-QXS/ansible_collections/fedora/linux_system_roles/roles/selinux/tasks/main.yml': line 41, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


    - name: Set permanent SELinux state if enabled
      ^ here
```

The Ansible file parser reads in the file, then checks for the presence of any modules referenced in the file, then does the other processing.  This happens *before* processing any `when` conditions.  So the effect is that if there is a module referenced in a file, that module must be present, even if the module will never actually be used at runtime due to conditionals.  So we may have to install the collection even if we don't use it.

Another way to do it would be to put the tasks that use those modules in separate task files e.g. have task files tasks/set_selinux_state.yml and tasks/set_selinux_state_el7.yml - then tasks/main.yml would use include_tasks: set_selinux_state_el7.yml if el7, or include_tasks: set_selinux_state.yml otherwise.  This is kinda ugly, but should get around the file parsing module problem.

spetrosi:
[citest]

richm:
still need to install fedora.linux_system_roles even on el7

spetrosi:
> Another way to do it would be to put the tasks that use those modules in separate task files e.g. have task files tasks/set_selinux_state.yml and tasks/set_selinux_state_el7.yml - then tasks/main.yml would use include_tasks: set_selinux_state_el7.yml if el7, or include_tasks: set_selinux_state.yml otherwise. This is kinda ugly, but should get around the file parsing module problem.

I did this by including modules not with FQDN but with a short name. This is less code and more readable. Also, with modules that exist in ansible 2.9 by default, like ini_files, this eliminates the need to have separate tasks for EL7 and notEl7 because ansible would pick the module available by the short name. EL7 picks the built-in ini_files, ansible-core picks ini_files from community.general.
The disadvantage is that it violates ansible's policy about using FQMN, which we anyway violate in other places. 

And with the short module name, Ansible doesn't fail on referencing non-existing modules.
WDYT?

> still need to install fedora.linux_system_roles even on el7

Right, opened https://github.com/linux-system-roles/tft-tests/pull/104

richm:
[citest]

tests/tests_all_transitions.yml
```diff
@@ -44,8 +44,10 @@
           ansible_facts['distribution_major_version'] is version('8', '>=') and
           ansible_version.full is version('2.10', '>=') }}"
 
+    # Bare name resolves to the ansible.posix builtin on EL7 (ansible 2.9) and
+    # redirects to the latest ansible.posix.selinux on newer systems.
     - name: Disable selinux and set kernel
-      ansible.posix.selinux:
+      selinux:  # noqa fqcn
```
richm, line 50: one problem is that I think we have to use `selinux_el7` on el7 and `selinux` on other platforms.

The next problem will be that `selinux_el7` is not in the default module lookup path.  I had a similar problem with the firewall role tests_ansible.yml which uses the module defined in the role directly - https://github.com/linux-system-roles/firewall/blob/main/tests/tests_ansible.yml#L26 - in order to solve this I created a directory tests/library and put a symlink to the role module there.

spetrosi: You mean this for is for the case when we'll write a test that calls selinux_el7 module?

richm: > You mean this for is for the case when we'll write a test that calls selinux_el7 module?

Yes.

spetrosi:
[citest]

richm (approved):

spetrosi:
[citest]

PR #375, opened 2026-08-21
fix: Relax ansible.posix and community.general version constraints [citest_skip]

spetrosi:
Enhancement: Relax ansible.posix and community.general version constraints

Reason: The caps (ansible.posix <2.2.0, community.general <12.0.0) existed to preserve EL7 compatibility, but we now vendor EL7 modules, hence other distros can use latest versions.

Result: The role runs on EL7 managed nodes natively, and users can install latest community collections versions.

### DIFF
--- a/meta/collection-requirements.yml
+++ b/meta/collection-requirements.yml
@@ -2,7 +2,5 @@
 ---
 collections:
   - name: ansible.posix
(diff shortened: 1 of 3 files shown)


spetrosi:
[citest]

PR #377, opened 2026-08-31
fix: Select SELinux modules by managed node python version

spetrosi:
Enhancement: Select the SELinux modules by the managed node's python version instead of the ansible version. Gate each task on `ansible_facts.python_version`:
- `ansible.posix` modules (`selinux`, `seboolean`): the vendored `*_python_27` module on python 2, and the bare name (redirecting to `ansible.posix`) on python 3.
- `community.general` modules (`sefcontext`, `selogin`): the vendored `*_python_36` module when python < 3.7, and the bare name (redirecting to `community.general`) when python >= 3.7.

Rename the vendored modules from `*_ansible_29` to `*_python_27`/`*_python_36` to reflect the latest python they back, gather the `python_version` fact subset, and update the tests accordingly.

Reason: The latest `community.general` modules use `from __future__ import annotations`, which requires python >= 3.7 and fails to import on managed nodes running python 3.6, regardless of the controller's ansible version. The vendored modules import only core `ansible.module_utils`, carry no collection dependency, and retain python 2 support, so they also cover ansible 2.9 hosts (e.g. EL7) where the collections cannot be installed. A fully-qualified name is avoided because an unresolvable FQCN aborts play parsing on ansible 2.9 even for a task skipped by `when`, so the bare name is used.

Result: The role runs the vendored `*_python_27`/`*_python_36` modules on old-python managed nodes (py2.7/py3.6, e.g. EL7/EL8) and the latest collection modules on nodes with newer python.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * SELinux task selection now adapts to the managed node’s Python version.
  * Improved compatibility for Python 2 and older Python 3 environments.
  * SELinux transition tests now use Python-version-based module selection.

* **Documentation**
  * Updated module names and usage examples to match the supported Python compatibility variants.

### DIFF
--- a/library/seboolean_python_27.py
+++ b/library/seboolean_python_27.py
@@ -12,7 +12,7 @@ __metaclass__ = type
 
 DOCUMENTATION = r"""
 ---
-module: seboolean_ansible_29
+module: seboolean_python_27
 short_description: Toggles SELinux booleans
 description:
      - Toggles SELinux booleans.
@@ -49,7 +49,7 @@ author:
 
 EXAMPLES = r"""
 - name: Set httpd_can_network_connect flag on and keep it persistent across reboots
-  seboolean_ansible_29:
+  seboolean_python_27:
     name: httpd_can_network_connect
     state: true
     persistent: true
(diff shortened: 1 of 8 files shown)


richm:
[citest]

spetrosi:
[citest]

richm (approved):

PR #387, opened 2026-09-04
fix: Move SELinux module calls to shared include task files

spetrosi:
Enhancement: Move the vendored and ansible.posix/community.general selinux, seboolean, sefcontext and selogin calls into shared include_*.yml files that main.yml pulls in with include_tasks, and update tests_all_transitions.yml to reuse include_selinux_state.yml via include_role with tasks_from.
Reason: An ansible.posix/community.general FQCN in a statically parsed task file aborts play parsing on ansible 2.9 even when the task is skipped, which previously forced bare module names with an fqcn lint suppression.
Result: The collection modules are now referenced by FQCN and only parsed at runtime, keeping ansible 2.9 support while dropping the bare-name workaround.

## Summary by CodeRabbit

* **Compatibility**
  * Improved SELinux state management across systems using different Python versions.
  * Added consistent handling for SELinux booleans, file contexts, login settings, and policy state across legacy and modern environments.

* **Refactor**
  * Streamlined SELinux operations with automatic selection of compatible implementations.
  * Preserved existing configuration inputs, enable/disable behavior, policy reload notifications, and operation results.
  * Updated transition tests to use the unified SELinux state workflow.


richm:
[citest_all]

spetrosi:
[citest_all]

tasks/include_seboolean.yml
```diff
@@ -0,0 +1,22 @@
...
+  seboolean_python_27: "{{ __selinux_seboolean_args }}"  # noqa args[module]
+  when:
+    - ansible_facts['python_version'] is version('3', '<')
+
+# Bare module name (not FQCN): an unresolvable FQCN aborts play parsing on
+# ansible 2.9 even for a task skipped by "when", so the bare name is used.
+# This runs only on managed nodes with python 3, where the bare name resolves
+# to the latest ansible.posix.seboolean.
+- name: Set SELinux booleans
+  seboolean: "{{ __selinux_seboolean_args }}"  # noqa fqcn args[module]
```
richm, line 15: @spetrosi I thought it had to use FQCN?  If you do have to use FQCN, then you'll have to use three include files for each module - include_seboolean.yml, include_seboolean_py27.yml, and include_seboolean_fqcn.yml - or something like that

spetrosi: It doesn't have to, I wanted it because it's cleaner in the upstream to have `ansible.posix.seboolean`, but kept the short-name as a workaround for ansible 2.9.

spetrosi:
[citest_all]

richm (approved):

PR #388, opened 2026-09-16
feat: Add an option to ignore the SELinux module cache when installing a module

FLAK-ZOSO:
Enhancement:

Add an optional `ignore_module_cache` setting to `selinux_modules`. When enabled, the role calls `semanage_set_ignore_module_cache()`, equivalent to `semodule --ignore-module-cache` (`semodule -C`), before installing the module.

Example:

```yaml
selinux_modules:
  - path: localmodule.pp
    state: enabled
    ignore_module_cache: true
```

The option defaults to `false`, so no backwards compatibility is lost.

Reason:

It is an overall positive thing to be able to configure this setting, however I find the specific need for it for what follows.

On affected EL8 systems, libsemanage can fail while installing a module from the role's temporary absolute path when its module cache is inconsistent:

```
libsemanage.semanage_direct_extract: Unable to access .../active/modules/400/<module>: No such file or directory
Problems processing filecon rules
Failed post db handling
Installing the same policy with semodule -C succeeds. 
```

While the [BZ 2010450](https://bugzilla.redhat.com/show_bug.cgi?id=2010450) was closed for lack of information, I see this happening on ~10% of my systems and I fail to identify a pattern; as the workaround so far has been to manually install the module passing the `--ignore-module-cache` (`-C`) option, I figured that having it as a feature in Ansible would be a better way to go.

Result:

Users can opt into rebuilding cached CIL files during module installation. Documentation and the existing checksum/idempotence test cover the new setting.

Signed-off-by: Mattia Marchese <mattia.marchese@taltech.ee>

## Summary by CodeRabbit

* **New Features**
  * Added the `ignore_module_cache` option for SELinux module installation. When enabled, modules are recompiled instead of using cached files; the option is disabled by default.
  * If the installed SELinux management library does not support this option, installation reports that it is unavailable.
* **Documentation**
  * Documented the option, its default behavior, and an example of its use.


tests/tests_selinux_modules_checksum.yml
```diff
@@ -25,7 +25,8 @@
       when: __run_selinux_modules_checksum
       vars:
         selinux_modules:
-          - {path: "selinux_modules/linux-system-roles-selinux-test-a.pp"}
+          - {path: "selinux_modules/linux-system-roles-selinux-test-a.pp",
+             ignore_module_cache: true}
```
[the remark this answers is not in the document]
FLAK-ZOSO: I guess it is time to ask another clanker to do what one clanker asks me to do; dead Internet theory.
I will be back in a moment.

richm:
[citest_all]

tests/tests_selinux_modules_checksum.yml
```diff
@@ -25,8 +25,7 @@
       when: __run_selinux_modules_checksum
       vars:
         selinux_modules:
-          - {path: "selinux_modules/linux-system-roles-selinux-test-a.pp",
-             ignore_module_cache: true}
+          - {path: "selinux_modules/linux-system-roles-selinux-test-a.pp"}
```
richm: That is - it has a python unit test, but it is not being tested in integration testing.  Will it cause problems to have it tested here in the integration test?

FLAK-ZOSO: You are right, no harms in leaving it there too.

The issue is that since the very moment the CI tests showed up as failed, clicking on them lead me to a 404 page, thus making it borderline impossible for me to understand what is wrong with them. Any clue on how I can explore the failures?

richm:
[citest_all]

tests/tests_selinux_modules_checksum.yml
richm: We had some issues with CI yesterday - I think they have been fixed - trying now

richm:
@bachradsusi this lgtm - do you see any issues?

bachradsusi:
I have no objection to the code.

I'd personally rather provide an integration test instead of the unit test. Something around these commands could work:

```
# semodule -i linux-system-roles-selinux-test-a.pp

# date > /var/lib/selinux/targeted/active/modules/400/linux-system-roles-selinux-test-a/cil

# semodule -i linux-system-roles-selinux-test-c.pp 
Symbol not inside parenthesis at line 1 of /var/lib/selinux/targeted/tmp/modules/400/linux-system-roles-selinux-test-a/cil
libsemanage.semanage_load_files: Error while reading from file /var/lib/selinux/targeted/tmp/modules/400/linux-system-roles-selinux-test-a/cil. (No data available).
semodule:  Failed!

# semodule -C -i linux-system-roles-selinux-test-c.pp

# 
```

I meant an equivalent playbook to these commands

tests/tests_selinux_modules_ignore_cache.yml
```diff
@@ -0,0 +1,108 @@
...
+        - name: Install module used to populate the cache
+          include_tasks: tasks/run_role_with_clear_facts.yml
+          vars:
+            selinux_modules:
+              - path: selinux_modules/linux-system-roles-selinux-test-a.pp
+
+        - name: Set module cache path
+          set_fact:
+            __selinux_test_module_cache: "{{ __selinux_test_store_root ~
+              '/targeted/active/modules/400/' ~
```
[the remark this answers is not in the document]
richm: I think this is ok.  Other tests assume `targeted` host.

richm (approved):

